PR #71, opened 2023-04-13
TypeScript migration

FunctionDJ:


src/pages/match/[matchId].tsx
```diff
@@ -0,0 +1,158 @@
...
+          </p>
+          <p className={styles.segment}>Income: {playerInMatch.properties}</p>
+        </div>
+      </div>
+    </div>
+  );
+};
+
+export default function Match() {
+  const [players, setPlayers] = useState<PlayerState | null | undefined>(null);
```
Comment (FemboyDeveloper) [inline src/pages/match/[matchId].tsx:75]: what exactly are we doing with this useState? definining it as PlayerState or null or undefined?

Comment (FunctionDJ) [reply src/pages/match/[matchId].tsx:75]: this looks like this temporarily. null or undefined is just because there's a quirk somewhere in the backend code where this is a possible value. later on it should be just players, no null or undefined.

PR #73, opened 2023-04-17
Home page update

FemboyDeveloper:
### Home Page
just added a very simple homepage so our users are greeted by something that looks nice to the eyes. 

### Setup your own local environment
Both README.md and CONTRIBUTING.md now contain a way to easily setup your own local environment plus some minor wording updates.


FemboyDeveloper:
@FunctionDJ if you wonder, I checked out your branch in order to preserve the whole example.env to .env stuff. Let me know if this is an issue for you, but this PR will make it smoother for the first time someone sets up their environment/update our README/CONTRIBUTE.md

src/styles/index.scss
```diff
@@ -0,0 +1,44 @@
+
+
+@use 'src/styles/variables';
+
+@import url('https://fonts.googleapis.com/css2?family=Poppins:wght@200&display=swap');
```
FunctionDJ, line 5: loading fonts from google servers can be a GDPR violation in EU countries. it's better if load it from our own server.

FunctionDJ (commented):
it would be better if we didn't use the google fonts CDN. (see code comment)

PR #81, opened 2023-04-19
Enable Tailwind CSS

WesKrug:
Now presenting the homepage we all know and love... but now with 100% more tailwind!
image

Included in the PR:

- Added Tailwind config! Now you can apply css classes on the fly so now instead of having to create a css class and apply it to your element like:

```
<div className="someClass"></div>
...
.someClass {
   padding: 2rem;
}
```

Now you can do the same thing directly on the div! Like so:
```
<div className="@p-8"></div>
```

- Added some default styles taken from our existing scss files. You can find the default in `/src/styles/globals.css` These are not final.


FemboyDeveloper (approved):
Everything seems fine, love that you re-made the homepage in tailwind!

PR #82, opened 2023-04-19
Migrating layout from demo

leekahung:
Migrated some stuff from the Demo.

Included Header and Footer from the Demo and created a new container called Layout. Footer can be toggled to be turned off depending on the page pages (could do the same for Header, if necessary).

Also made some headway on the mapGrid to be dynamic by tweaking the mapTiles


WesKrug:
@leekahung so I just added tailwindcss to the repo with the hopes of reducing the number of stylesheets we have.

I'd like to stay away from using raw scss files. What are your thoughts?

leekahung:
Oh, I don't mind @WesKrug. I haven't really dabbled with tailwind before, so I had to learn it, but reducing the amount of stylesheets are something I can get behind.

leekahung:
Moving branch from fork to main repo.

PR #83, opened 2023-04-20
Migrating layout from demo 2

leekahung:
Same as before in PR #82, but I moved the branch from my fork to the main repo.

Although there are changes related to the mapGrid, I think we can at least push into the main branch the changes with regards to a Layout component (with Header/Navbar and Footer). 

Link/Images for navbar could be altered depending on what we would want later on.


leekahung:
Going to swap to tailwind after potential merge now that it's added to the project.

leekahung:
Implementing Match Page idea to PR.

match_page_layout_idea

As you can see below, the smallest I could get it to go at this moment with the existing design is 330px by 722px. Things starts breaking after that. But this is only optimized for the Match Page.

Screen Shot 2023-04-20 at 3 21 36
Screen Shot 2023-04-20 at 3 22 34
Screen Shot 2023-04-20 at 3 23 27
Screen Shot 2023-04-20 at 3 23 45
Screen Shot 2023-04-20 at 3 24 06
Screen Shot 2023-04-20 at 3 24 37

FemboyDeveloper:
@WesKrug is this ready for merging?

FemboyDeveloper:
I just checked this out, looks pretty nice overall! Its a nice skeleton!

src/components/Footer.tsx
```diff
@@ -0,0 +1,37 @@
...
+        <Link className="@text-base-a" href="/">
+          Terms of Use
+        </Link>
+        <Link className="@text-base-a" href="/">
+          Rules
+        </Link>
+      </nav>
+
+      <nav className="@flex @justify-center @gap-5">
+        <img className="footerLogo" src={rLogoPath} alt="Reddit Logo" />
```
WesKrug, line 26: Any reason you're not just using @h-8 for 32px here?

src/styles/template/layout.scss
```diff
@@ -0,0 +1,104 @@
...
+        height: 60px;
+      }
+
+      .headerLogo {
+        max-height: 50px;
+      }
+    }
+  }
+
+  footer {
```
WesKrug, line 97: This appears to be unused

src/pages/home.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export default function HomePage() {
+  return (
+    <section>
+      <Head>
+        <title>Wars World Home Page</title>
+        <link rel="icon" href="/favicon.ico" />
+      </Head>
+
+      <div className="@rounded-lg @bg-black/70 @center @p-8 @max-w-3xl isProgress">
```
WesKrug, line 11: Super nitpicky. should be `inProgress`, not `isProgress`
```suggestion
      <div className="@rounded-lg @bg-black/70 @center @p-8 @max-w-3xl inProgress">
```

src/styles/template/layout.scss
```diff
@@ -0,0 +1,104 @@
...
+    overflow: scroll;
+    -ms-overflow-style: none;
+    scrollbar-width: none;
+  }
+
+  main::-webkit-scrollbar {
+    display: none;
+  }
+
+  .isProgress {
```
WesKrug, line 33: same nitpick
```suggestion
  .inProgress {
```

WesKrug (approved):

leekahung:
Player Box updates looks like this. Last image shows a working burger menu for mobile.

Screen Shot 2023-04-21 at 13 35 50
Screen Shot 2023-04-21 at 13 28 03
Screen Shot 2023-04-21 at 13 28 25
Screen Shot 2023-04-21 at 13 33 26

src/pages/home.tsx
leekahung: Got it, getting this fixed. It appears I've missed the className in the spreadsheet as well (Line 33 of layout.scss). I'll change it there as well.

src/styles/template/layout.scss
leekahung: Fixed.

src/components/Footer.tsx
leekahung: Good spot. No reason in particular. I should just use @h-8 from tailwind.

src/styles/template/layout.scss
leekahung: It's being used for footer height: 160px

src/pages/match/[matchId].tsx
```diff
@@ -26,55 +34,71 @@ const PlayerBox = ({
...
+          />
+          <img
+            className="@absolute @h-8 @top-1 @right-1 @bg-slate-200"
+            src={`/img/nations/${playerInMatch.nation}.webp`}
+          />
+        </div>
+        <div className="playerNationBox">
+          <div className="playerUsernameIconAndIngameStats">
+            <div className={`playerUsernameAndIcon`}>
+              <span>{playerInMatch.username}</span>
```
Comment (leekahung) [inline src/pages/match/[matchId].tsx:71]: This is just a singular element now with just playerUsername. No need to wrap in div.

src/styles/grid/global.scss
```diff
@@ -8,8 +8,140 @@
...
+        display: grid;
+        grid-template-columns: 200px;
+
+        @media (max-width: 1000px) {
+          grid-template-columns: none;
+          grid-template-rows: repeat(3, 40px);
+        }
+
+        .playerUsernameAndIcon {
+          display: flex;
```
leekahung, line 69: Can migrate some props from all classes under playerBox (this class is a good example)

src/styles/grid/global.scss
```diff
@@ -8,8 +8,140 @@
...
+      width: clamp(6rem, 2.59rem + 12.07vw, 9rem);
+      font-size: clamp(0.875rem, 0.75rem + 0.625vw, 1.5rem);
+      background-color: lightgrey;
+      border-radius: 10px;
+      padding: 2px;
+    }
+  }
+
+  @media (max-width: 700px) {
+    flex-direction: column;
```
leekahung, line 144: Redundany, flex-direction never changes for gameInnerBox

leekahung (commented):
Minor check going through every file. Please review.

src/styles/grid/global.scss
```diff
@@ -438,8 +579,7 @@ h1 {
   padding: 0.5em;
 }
 .playerBox {
-  width: 100%;
-  z-index: 25;
+  z-index: 20;
```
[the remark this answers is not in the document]
leekahung: Resolved redundancy

src/pages/match/[matchId].tsx
Comment (leekahung) [reply src/pages/match/[matchId].tsx]: Added tailwind classes for playerUsername, playerIngameExp, playerIngameInfo, gameTime, and others inside playerBox

src/styles/grid/global.scss
leekahung: Resolved, have changed props

src/styles/grid/global.scss
leekahung: Lines removed.

FemboyDeveloper:
@leekahung is this ready for merging?

src/components/layout.tsx
```diff
@@ -0,0 +1,51 @@
...
+import Navbar from './Navbar';
+import Footer from './Footer';
+import { useMediaQuery } from 'utils/useMediaQuery';
+
+interface Props {
+  addFooter?: boolean;
+  children: JSX.Element;
+}
+
+export default function Layout({ addFooter, children }: Props) {
```
FunctionDJ, line 11: please use named exports when possible (next pages must use default exports).
they are better for refactoring. more info: https://stackoverflow.com/a/33307487

src/pages/home.tsx
FunctionDJ, line 1: pages should not be used as components in other files (`index.tsx`). it should either be a page or a component (components folder) which then doesn't get rendered as it's own page.
as it is right now we'd have `/` and `/home` (two pages)  with the same or similar content.

src/pages/index.tsx
```diff
@@ -1,25 +1,23 @@
-import Head from 'next/head';
-
+import Layout from 'components/layout';
+import HomePage from './home';
 
+export default function IndexPage() {
+  const documentHeight = () => {
```
FunctionDJ, line 5: see my other comment about javascript-less responsive design.
but aside from that `documentHeight` would be a good name for a value but not for a function. if we kept this function, `updateDocumentHeightCSSProperty` or something similar would be better.

src/pages/match/[matchId].tsx
```diff
@@ -13,68 +14,97 @@ import {
...
+import IngameInfo from 'components/IngameInfo';
 
-const smallPadding: CSSProperties = {
-  padding: '0.2rem',
-};
+interface INationColors {
+  [key: string]: string;
+}
+
+interface INationColorGradients {
```
Comment (FunctionDJ) [inline src/pages/match/[matchId].tsx:25]: both of these interfaces express the same as `Record<string, string>`

src/pages/match/[matchId].tsx
```diff
@@ -13,68 +14,97 @@ import {
...
   playerInMatch: PlayerInMatch;
-}) => {
+}
+
+const PlayerBox = ({ playerTurn, playerInMatch: playerInMatch }: Props) => {
   const time = new Date(0);
   time.setSeconds(playerInMatch.timePlayed ?? 1);
+  console.log(playerTurn);
+
+  const nationColors: INationColors = {
```
Comment (FunctionDJ) [inline src/pages/match/[matchId].tsx:39]: objects that don't depend on component data should be created outside of components in order not to trigger unnecessary re-renders.

src/pages/match/[matchId].tsx
```diff
@@ -13,68 +14,97 @@ import {
...
+          } playerCOBox`}
         >
-          <p className={styles.segment}>Units: {playerInMatch.unitCount}</p>
-          <p className={styles.segment}>
-            Army-Value: {playerInMatch.properties * 1000}
-          </p>
-          <p className={styles.segment}>Income: {playerInMatch.properties}</p>
+          <img
+            className={`@absolute @bottom-0 playerCOIcon ${
+              playerTurn ? null : 'isNotPlayerTurn'
```
Comment (FunctionDJ) [inline src/pages/match/[matchId].tsx:63]: `null` gets stringified to `"null"`, not to empty string `""`!
so please use empty string :)

src/pages/match/[matchId].tsx
```diff
@@ -224,87 +256,130 @@ export default function Match() {
     onData: console.log,
   });
 
-  const makeMove = trpc.match.makeMove.useMutation();
+  // Original functionality for turns
+  // const makeMove = trpc.match.makeMove.useMutation();
+
+  // mock functionality for testing css transitions
+  const [turn, setTurn] = useState({ player1: true, player2: false });
```
Comment (FunctionDJ) [inline src/pages/match/[matchId].tsx:263]: this object reference creation on-render can be unsafe in certain conditions. if you must use objects in this way and they are independent of components state like in this case, move them outside the component so the reference stays the same.

src/styles/grid/global.scss
```diff
@@ -8,8 +8,136 @@
...
+    .playerCOAndNationBox {
+      flex-direction: row-reverse;
+    }
+  }
+}
+
+.gameInnerBox {
+  flex-direction: column;
+
+  .gameTime {
```
FunctionDJ, line 125: you wouldn't want to re-use this class name `.gameTime` for anything else, so in my opinion these styles shouldn't be restricted to children of `.gameInnerBox`. the nesting suggests that the class name `.gameTime` might get different styles when used as a child of another specific class or id.

src/styles/match.module.css
```diff
@@ -1,20 +1,37 @@
...
+@media (max-width: 700px) {
+  .match {
+    flex-direction: column;
+    align-items: center;
+  }
+}
+
+@media (max-width: 700px) and (max-height: 800px) {
+  .match {
+    gap: 10px;
```
FunctionDJ, line 29: thoughts about using `em` or `rem` over pixels?

src/styles/template/layout.scss
```diff
@@ -0,0 +1,100 @@
+:root {
+  --layoutGridRows: 100px 1fr 160px;
+  --doc-height: 100%;
+}
+
+#__next,
+html,
+body {
+  height: 100vh;
+  height: var(--doc-height);
```
FunctionDJ, line 10: idk about this one chief

src/styles/template/layout.scss
```diff
@@ -0,0 +1,100 @@
...
+  overflow: hidden;
```
FunctionDJ, line 11: this is a hack to hide scrollbars in this case because an element can be too big.
it'd be better to fix the issue causing elements to grow too large instead of patching it this way which can cause scrolling issues.

src/styles/template/layout.scss
```diff
@@ -0,0 +1,100 @@
...
+.layout {
+  grid-template-rows: var(--layoutGridRows);
+
+  main {
+    overflow: scroll;
+    -ms-overflow-style: none;
+    scrollbar-width: none;
+  }
+
+  main::-webkit-scrollbar {
```
FunctionDJ, line 29: hax?

src/utils/useMediaQuery.ts
```diff
@@ -0,0 +1,23 @@
+import { useState, useEffect } from 'react';
+
+export const useMediaQuery = (query: string) => {
+  const [matches, setMatches] = useState(false);
+
+  useEffect(() => {
+    const media = window.matchMedia(query);
+    if (media.matches !== matches) {
```
FunctionDJ, line 8: this if is effectively a no-op because the state contains a primitive value.

FunctionDJ (changes requested):
interesting stuff. some of my comments are picky and are more about learning about the decisions behind the code rather than the actual code, please excuse that.

leekahung:
@FemboyDeveloper Yeah, I think we're ready for mergin. Lemme go through the feedback from @FunctionDJ first and see what still need to be updated or modified.

.vscode/settings.json
```diff
@@ -1,12 +1,13 @@
 {
   "typescript.tsdk": "node_modules/typescript/lib",
-  "editor.defaultFormatter": "rvest.vs-code-prettier-eslint",
```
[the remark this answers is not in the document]
leekahung: It appears to be a VSCode thing. I didn't have the extension, so it originally showed as a value not accepted.

Found the extension and installed it locally. I should be able to turn this back on. and remove the only format with just prettier.

.vscode/settings.json
```diff
@@ -1,12 +1,13 @@
...
   "editor.formatOnSave": true,
   "editor.formatOnPaste": false,
   "editor.formatOnType": false,
   "editor.formatOnSaveMode": "file",
-  "vs-code-prettier-eslint.prettierLast": false,
   "[prisma]": {
     "editor.defaultFormatter": "Prisma.prisma"
+  },
+  "[typescriptreact]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
```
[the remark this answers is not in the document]
leekahung: Found where the issue was. Should be able to revert back to the original settings.

src/components/Footer.tsx
```diff
@@ -0,0 +1,47 @@
...
+          target="_blank"
+          rel="noreferrer"
+        >
+          <img className="@h-8" src={gLogoPath} alt="GitHub Logo" />
+        </Link>
+      </nav>
+
+      <p className="@text-center @text-base-p @p-0 @mx-1">
+        Advance Wars is (c) 1990-2001 Nintendo and (c) 2001 Intelligent Systems.
+        All images are copyright their respective owners.
```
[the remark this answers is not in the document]
leekahung: Fixed "All images are copyright their respective owners" to "All images are copyright of their respective owners"

public/img/layout/burgerMenu.png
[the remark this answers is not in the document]
leekahung: Originally, I had an icon lying around so I wanted to use it. Other than that, there shouldn't been any issues swapping to an HTML-based one.

I'll change it.

src/components/IngameInfo.tsx
```diff
@@ -0,0 +1,17 @@
+interface Props {
+  iconPath: string;
+  value: string | number;
```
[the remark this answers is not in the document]
leekahung: Yes, "value" should be sufficient for mocking. We could name it "ingameStat" for future-proofing. That should be a better name.

As for the typing, I don't really know what type of data that's going to be passed into it yet, just that it might be a number (i.e. income amount, number of units, etc.) or a time (i.e. player turn timer).

I could use a Date for typing the time, but it's being passed as a string in the mock, so the typing is left like that.

src/styles/match.module.css
leekahung: That was the initial idea, which was why gap for .match is set in rem. But It's difficult to maintain a "min-gap" that takes into account media queries. There are many times where rem makes the component stick past the screen because the screen was not small enough in either the vertical or horizontal directions.

src/styles/template/layout.scss
leekahung: I mean, I could always use dvh. But I don't think dvh is supported by FireFox? [https://caniuse.com/?search=dvh](https://caniuse.com/?search=dvh)

I need a fallback regardless. [see this stackoverflow](https://stackoverflow.com/questions/37112218/css3-100vh-not-constant-in-mobile-browser#:~:text=Yes%2C%20100vh%20is%20100%25%20of,only%20to%20this%20parent%20height).)

src/styles/template/layout.scss
leekahung: No, I want to hide the scroll bar completely for a better experience.

src/styles/template/layout.scss
leekahung: No, I just want to hide the scroll bar completely.

src/utils/useMediaQuery.ts
leekahung: I'm not quite understanding this feedback?

src/components/NavItem.tsx
```diff
@@ -0,0 +1,28 @@
+import Link from 'next/link';
+import { useMediaQuery } from 'utils/useMediaQuery';
+
+interface Props {
+  text: string;
+  location: string;
+}
+
+export default function NavItem({ text, location }: Props) {
+  const query500 = useMediaQuery('(max-width: 500px)');
```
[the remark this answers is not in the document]
leekahung: No, it's not required for tailwind from what I've saw in the docs. But we didn't not set up our existing tailwind config to handle responsive breakpoints, so I need a custom hook for now.

src/components/Navbar.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const awLogoPath = '/img/layout/awLogo.webp';
+  const burgerMenuPath = '/img/layout/burgerMenu.png';
+  const query500 = useMediaQuery('(max-width: 500px)');
+  const [showLinks, setShowLinks] = useState(false);
+
+  return (
+    <header className="@w-screen @relative @z-30">
+      <nav className="@flex @justify-between @items-center @bg-gray-800 @mx-auto @px-3 navHeader">
+        <Link href="/">
+          <img className="@flex headerLogo" src={awLogoPath} alt="AW Logo" />
```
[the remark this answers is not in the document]
leekahung: Could switch to it. Didn't now about this component in NextJS.

src/components/layout.tsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import Navbar from './Navbar';
+import Footer from './Footer';
+import { useMediaQuery } from 'utils/useMediaQuery';
+
+interface Props {
+  addFooter?: boolean;
```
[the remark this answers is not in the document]
leekahung: Can be done. Changing prop from addFooter to just footer

src/components/layout.tsx
```diff
@@ -0,0 +1,51 @@
...
+  children: JSX.Element;
+}
+
+export default function Layout({ addFooter, children }: Props) {
+  const query800H = useMediaQuery('(max-height: 800px)');
+
+  if (typeof document !== 'undefined') {
```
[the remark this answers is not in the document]
leekahung: Yes, it's a check. I can swap to window instead of document.

src/components/layout.tsx
```diff
@@ -0,0 +1,51 @@
...
+    if (addFooter) {
+      if (query800H) {
+        document.documentElement.style.setProperty(
```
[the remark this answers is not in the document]
leekahung: Usually, it's pretty easy to select from :root for dynamically adjusting dynamic CSS properties outside of SASS [see MDN link](https://developer.mozilla.org/en-US/docs/Web/CSS/:root)

However, Next seems to be doing its own thing?

src/pages/_app.tsx
```diff
@@ -1,4 +1,6 @@
-import '../styles/globals.css'
+import '../styles/globals.css';
+import '../styles/grid/global.scss';
```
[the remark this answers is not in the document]
leekahung: Moved both SASS files into global.css (now renamed global.scss) for importing by using @use

src/components/layout.tsx
```diff
@@ -0,0 +1,51 @@
...
+    }
+  }
+
+  return (
+    <div className="@grid @relative @h-full @w-screen layout">
+      <Navbar />
+      <main className="@flex @h-full @w-screen @justify-center @items-center">
+        {children}
+      </main>
+      {addFooter ? <Footer /> : null}
```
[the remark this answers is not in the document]
leekahung: Got this changed

src/pages/home.tsx
```diff
@@ -0,0 +1,37 @@
+import Head from 'next/head';
+
+export default function HomePage() {
+  return (
+    <section>
+      <Head>
+        <title>Wars World Home Page</title>
+        <link rel="icon" href="/favicon.ico" />
```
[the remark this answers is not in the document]
leekahung: Moving this to _app.tsx

src/pages/home.tsx
```diff
@@ -0,0 +1,37 @@
...
+        <h1>Welcome to Wars World!</h1>
+        <p>First of all, congrats on setting your developer environment!</p>
+        <p>
+          We are in the process of refactoring &quot;AW-Competitive&quot;/Wars
+          World demo (FemboyDeveloper, Rey, Yeti et al.) into
+          Typescript/React/Next.js + tRPC + Prisma/SQL (FunctionDJ&apos;s idea)
+        </p>
+        <p>
+          Before you decide to take on one of our{' '}
+          <a href={'https://github.com/WarsWorld/WarsWorld/issues'}>
```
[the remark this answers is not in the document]
leekahung: You are right.

This was a very silly mistake on my part. I didn't know why I use that instead of a string, lol.

src/pages/howtoplay/index.tsx
```diff
@@ -0,0 +1,9 @@
+import Layout from 'components/layout';
+
+export default function howToPlayPage() {
```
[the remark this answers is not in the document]
leekahung: I didn't know NextJS pages are React components. Noted and changed it to PascalCase.

src/components/layout.tsx
leekahung: Changed Layout as a named export. I should change existing components to named exports as well.

src/pages/home.tsx
leekahung: Should I move home and about in their own folder in pages? I'm a little confused over this feedback.

src/pages/index.tsx
leekahung: Noted, I'll use updateDocumentHeightCSSProperty as the function name.

src/pages/match/[matchId].tsx
Comment (leekahung) [reply src/pages/match/[matchId].tsx:25]: I'll just keep INationColors instead of having both.

src/pages/match/[matchId].tsx
Comment (leekahung) [reply src/pages/match/[matchId].tsx:39]: Remove console.log from Line 37?

src/pages/match/[matchId].tsx
Comment (leekahung) [reply src/pages/match/[matchId].tsx:63]: I've noticed that and had been swapping them out. I guess I missed this one. Changing it to an empty string.

src/pages/match/[matchId].tsx
Comment (leekahung) [reply src/pages/match/[matchId].tsx:263]: I'm a little confused on this feedback? Do you mean to lift this to a global state?

src/styles/grid/global.scss
leekahung: Moving .gameTime outside of .gameInnerBox

src/styles/template/layout.scss
leekahung: No, just want to hide the scroll bar as intended.

src/pages/match/[matchId].tsx
Comment (FunctionDJ) [reply src/pages/match/[matchId].tsx:25]: actually that type doesn't do anything because it only gets used once and doesn't convey any meaningful information. i suggest to just remove it. typescript will still infer the object shape.

src/utils/useMediaQuery.ts
FunctionDJ: when you set a state, react will only re-render if the new value is a different primitive value or a different object reference.
with this `if` you're preventing the case where you would set the same primitive value that's currently set, but that wouldn't cause react to re-render.
so you can and should remove this `if` because it effectively makes no difference.

src/pages/match/[matchId].tsx
Comment (leekahung) [reply src/pages/match/[matchId].tsx:39]: Removed nationsColors

src/pages/match/[matchId].tsx
Comment (leekahung) [reply src/pages/match/[matchId].tsx:263]: It's just there for mocking the pass turn functionality. We can leave it for now.

src/utils/useMediaQuery.ts
leekahung: I see what you mean.

It does need the condition though for the listener to pick up the media query, so I moved the if statement inside listener, replacing what it was there originally.

src/pages/match/[matchId].tsx
Comment (leekahung) [reply src/pages/match/[matchId].tsx]: Oh! I get was you mean now.

Okay, I'll remove the interface in favor of a generic Record.

src/styles/template/layout.scss
leekahung: Going to fall back on vh. If vh becomes insufficient as a fall back, we can re-consider --doc-height. For now, I'll leave them out.

src/pages/match/[matchId].tsx
Comment (leekahung) [reply src/pages/match/[matchId].tsx:247]: Actually, it doesn't even have to be an object with booleans. Thinking about it again, I should just straight up use a boolean.

If I do const [turn, setTurn] = useState(true)

I should just pass turn as a prop to player 1 and !turn to player 2.

Damn, even if it's just a mock, I probably over-complicated the problem. Lol.

src/pages/match/[matchId].tsx
```diff
@@ -14,18 +14,14 @@ import {
 import { trpc } from 'utils/trpc';
 import { Army } from 'utils/wars-world-types';
 import styles from '../../styles/match.module.css';
-import Layout from 'components/layout';
+import { Layout } from 'components/layout';
 import { useMediaQuery } from 'utils/useMediaQuery';
-import IngameInfo from 'components/IngameInfo';
+import { IngameInfo } from 'components/IngameInfo';
 
 interface INationColors {
```
Comment (FunctionDJ) [inline src/pages/match/[matchId].tsx]: if you must use a custom type name, at least do `type INationColors = Record<string, string>`... or like just get rid of the type because it's shape doesn't transfer any useful information.

src/pages/match/[matchId].tsx
```diff
@@ -34,16 +30,8 @@ interface Props {
...
-
-  const nationColors: INationColors = {
-    blue: '@bg-blue-400',
-    orange: '@bg-orange-400',
-    green: '@bg-green-400',
-    yellow: '@bg-yellow-400',
-  };
 
-  const nationColorGradients: INationColorGradients = {
+  const nationColorGradients: INationColors = {
```
Comment (FunctionDJ) [inline src/pages/match/[matchId].tsx]: this object should live outside the react component.

FunctionDJ (changes requested):

public/img/layout/burgerMenu.png
FunctionDJ, line 1: please remove if unused

FunctionDJ (changes requested):

src/components/layout.tsx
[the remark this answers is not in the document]
FunctionDJ: using `:root` in CSS is fine, but *writing* to the DOM using the DOM API from inside react for a media query workaround... that's what's surprising me.

src/components/NavItem.tsx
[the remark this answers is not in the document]
leekahung: Yeah, I can do that. I'll add the TODO comment around

src/pages/_app.tsx
```diff
@@ -1,9 +1,10 @@
-import '../styles/globals.css';
-import '../styles/grid/global.scss';
-import '../styles/template/layout.scss';
+import '../styles/globals.scss';
+// import '../styles/grid/global.scss';
```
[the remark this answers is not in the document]
leekahung: This has been updated in recent commits

public/img/layout/burgerMenu.png
leekahung: Removing unused png.

src/pages/_app.tsx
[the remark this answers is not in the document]
leekahung: Comment has already been removed in the updated file:

```
import '../styles/globals.scss';
import type { Session } from 'next-auth';
```

src/components/layout.tsx
```diff
@@ -45,7 +45,7 @@ export default function Layout({ addFooter, children }: Props) {
       <main className="@flex @h-full @w-screen @justify-center @items-center">
         {children}
       </main>
-      {addFooter ? <Footer /> : null}
+      {footer ? <Footer /> : null}
```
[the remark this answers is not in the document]
leekahung: Changing this from ternary to logical and

src/pages/match/[matchId].tsx
Comment (leekahung) [reply src/pages/match/[matchId].tsx]: Record<string, string> is in use in the updated version, original Interface has already been removed:

```
interface Props {
  playerTurn: boolean;
  playerInMatch: PlayerInMatch;
}

const PlayerBox = ({ playerTurn, playerInMatch: playerInMatch }: Props) => {
  const time = new Date(0);
  time.setSeconds(playerInMatch.timePlayed ?? 1);

  const nationColorGradients: Record<string, string> = {
    blue: '@bg-gradient-to-l @from-blue-400',
    orange: '@bg-gradient-to-l @from-orange-400',
    green: '@bg-gradient-to-l @from-green-400',
    yellow: '@bg-gradient-to-l @from-yellow-400',
  };
```

src/pages/match/[matchId].tsx
Comment (FunctionDJ) [reply src/pages/match/[matchId].tsx]: for mock you could do `useState<"player1" | "player">("player1")`

src/styles/template/layout.scss
FunctionDJ: i meant the two height properties.

src/pages/match/[matchId].tsx
Comment (leekahung) [reply src/pages/match/[matchId].tsx]: INationColors and INationColorGradients has been removed in favor of Record<string, string> (see Line 31 of the latest commit)

src/styles/template/layout.scss
FunctionDJ: why is there a scrollbar?

src/utils/useMediaQuery.ts
```diff
@@ -5,12 +5,11 @@ export const useMediaQuery = (query: string) => {
 
   useEffect(() => {
     const media = window.matchMedia(query);
-    if (media.matches !== matches) {
-      setMatches(media.matches);
-    }
 
     const listener = () => {
-      setMatches(media.matches);
+      if (media.matches !== matches) {
```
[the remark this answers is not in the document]
leekahung: Added TODO for when we switch from custom Hook to tailwind media queries.

But I believe you mean just to use setMatches(media.matches) under const media in Line 8?

src/components/layout.tsx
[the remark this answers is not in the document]
leekahung: Is there a way to dynamically change the variable in :root outside of this method? I thought this was normal? For example: [stackoverflow](https://stackoverflow.com/questions/37801882/how-to-change-css-root-color-variables-in-javascript)

src/pages/match/[matchId].tsx
Comment (leekahung) [reply src/pages/match/[matchId].tsx]: I could use a more complicated state, but I do think for the mock, a simple true/false is enough.

src/styles/template/layout.scss
leekahung: Overflow hidden removed in the updated file.

src/styles/template/layout.scss
leekahung: This is for edge cases like with wide screens, but small device heights like the Google Next Hub. Still needs scrolling for now, but needs to be fixed the future. I'll add a TODO here.

src/styles/template/layout.scss
leekahung: Oh, this is for CSS fallback.

Since it's CSS, if the property on the bottom fails, it'll always read the top one and if the bottom one doesn't fail, it'll always read the bottom one. It's a safe way to ensure the property is set based on browser compatibility.

src/components/IngameInfo.tsx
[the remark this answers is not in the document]
leekahung: I've also renamed iconPath as ingameStatIconPath for better clarity

src/pages/_app.tsx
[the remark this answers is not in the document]
leekahung: Removing line for the unused SASS files from here

src/pages/home.tsx
leekahung: Have moved home.tsx to components and renamed it Home.tsx

src/pages/index.tsx
leekahung: After thinking about it some more, we can remove this in favor of vh, dvh in css

src/pages/match/[matchId].tsx
Comment (leekahung) [reply src/pages/match/[matchId].tsx]: Removed interface completely in favor of Record

src/utils/useMediaQuery.ts
leekahung: I think I get what you mean now. Dropping conditional completely and just leave the function inside it in line 9

src/pages/match/[matchId].tsx
Comment (FunctionDJ) [reply src/pages/match/[matchId].tsx]: it's still not outside the component :)

FunctionDJ (approved):
there's this one really minor thing left (independent object in react component) but it's not worth dragging this out any longer.
LGTM.

src/pages/match/[matchId].tsx
Comment (leekahung) [reply src/pages/match/[matchId].tsx]: Oh, you meant the nationColorGradient object and not the types/interface! My mistake. I know what to do. I'll leave them outside of this PR for now, but make a note for it.

FemboyDeveloper:
Just checked it out, absolutely wonderful work, thanks for your extensive help FunctionDJ!

PR #88, opened 2023-04-25
Configuring Tailwind for media queries and SASS refactoring

leekahung:
Currently implementing media queries for Tailwind in tailwind.config.js

Migrating some CSS properties from SASS to Tailwind


src/styles/globals.scss
```diff
@@ -36,13 +36,13 @@
 
   p {
     font-size: 0.75rem;
-    font-size: clamp(0.75rem, 0.65rem + 0.5vw, 1.25rem);
+    font-size: clamp(0.75rem, 0.5682rem + 0.9091vw, 1.25rem);
```
megagreg72, line 39: lol wut
how did you come up with these numbers?

tailwind.config.js
```diff
@@ -1,21 +1,28 @@
...
+      phone: '320px',
+      tablet: '480px',
+      smallscreen: '768px',
+      laptop: '1024px',
+      desktop: '1200px',
+      short: { raw: '(min-height): 600px' },
+      medium: { raw: '(min-height: 700px)' },
+      tall: { raw: '(min-height: 750px)' },
+      verytall: { raw: '(min-height: 800px)' },
     },
```
megagreg72, line 25: I generally _try_ to avoid creating CSS for a bunch of different screen sizes, and use `flex` to grow/shrink regions instead if possible.  Ideally, that would prevent code duplication and allow the page to look nice even if the browser window is a weird size (not fullscreen).
However, if we want drastic formatting changes at certain screen-size thresholds it might make sense to do this.  I'm also not familiar with tailwind `screens` so maybe I'm missing something.

src/styles/globals.scss
leekahung: Oh, I'm using a calculator for those clamp functions. Set what I want as the min and max size and account for min and max screen size. [See link to functions](https://chriskirknielsen.com/blog/modern-fluid-typography-with-clamp/)

tailwind.config.js
leekahung: Most of the code is just running with tablet (480px ato 767px), smallscreen (768px to 1023px), and laptop (1024px to 1200px) for width breakpoints and medium (700px to 749px) and tall (750 to 749px) for height breakpoints. (See bottom of SASS  files where all the @screen rules are)

Additional breakpoints were simply set for a few edge cases.

I think we can scale down some of these as suggested and check which transformations are unneeded.

leekahung: See Line 493 in src/styles/grid/global.scss for an example of edge cases of phone, which was used once.

For some reason, Tailwind won't update the styling if I set padding directly to that class. I'm not sure why, but I would prefer not using @screen phone just for rare cases like this.

src/styles/globals.scss
leekahung: Actually, I might as well incorporate the function as a mixin. That would probably be better than what I'm doing now.

I'll make a TODO for it.

leekahung: Got this updated to use an internal mixin for clamp calculations. Thanks for the feedback!

tailwind.config.js
leekahung: Oh! I just found the issue that was causing issues. It seems like Tailwind needs to make sure both CSS classes are called properly if deeply nested. I think we won't need the extra breakpoints at the moment. I'll be leaving just:

- tablet (min-width: 480px)
- smallscreen  (min-width: 768px)
- desktop (min-width: 1024px)
- short  (min-height: 600px)
- medium  (min-width: 700px)

All other cases could be handled, perhaps though the custom useMediaQuery hook for transformations or a better alternative in the future.

leekahung:
Updating commits based on recent feedback to Tailwind config file and SASS styling.

src/styles/globals.scss
megagreg72: the new mixin looks great!

tailwind.config.js
megagreg72: nicely done! I didn't mean to imply we had completely redo everything to get rid screen-size breakpoints, just "use with caution".  your recent changes all look good

megagreg72 (approved):
Nicely done!  If someone else wants to review this they are welcome to, but we should probably review+merge sooner rather than later because PRs get harder to review the bigger they get. 

One trick I often do to make PRs easier to review is to make one PR for "code cleanup" (like splitting one function into two, or moving stuff to a different file) followed by another PR for "actual changes".  That way, the reviewer(s) can easily check that the first PR just moves stuff around and it is way easier to read the second PR.

megagreg72:
> 

however, in the early stages of a project there is lot of new stuff and a lot "changes" which are just getting rid of placeholders, so maybe that strategy is more trouble than it is worth atm 🤷‍♂️

megagreg72 (approved):

leekahung:
> Nicely done! If someone else wants to review this they are welcome to, but we should probably review+merge sooner rather than later because PRs get harder to review the bigger they get.
> 
> One trick I often do to make PRs easier to review is to make one PR for "code cleanup" (like splitting one function into two, or moving stuff to a different file) followed by another PR for "actual changes". That way, the reviewer(s) can easily check that the first PR just moves stuff around and it is way easier to read the second PR.

Good idea. I think I'll do that for future PRs. I do want to make upcoming PRs smaller and easier to digest.

leekahung:
> > 
> 
> however, in the early stages of a project there is lot of new stuff and a lot "changes" which are just getting rid of placeholders, so maybe that strategy is more trouble than it is worth atm 🤷‍♂️

Hopefully it wouldn't be too much of trouble. XD

FemboyDeveloper:
> Good idea. I think I'll do that for future PRs. I do want to make upcoming PRs smaller and easier to digest.

Yes please, It's a bit hard to code review when its one large PR over just one smaller, specific one.

src/components/Navbar.tsx
```diff
@@ -1,59 +1,38 @@
...
+import { NavLinks } from './NavLinks';
 import { useMediaQuery } from 'utils/useMediaQuery';
+import { NavLinksMobile } from './NavLinksMobile';
 
 export function Navbar() {
   const awLogoPath = '/img/layout/awLogo.webp';
-  // TODO: Hook can be removed after media query is implemented for tailwind
-  const query500 = useMediaQuery('(max-width: 500px)');
+  // Might keep useMediaQuery to conditionally switch between full navbar and hamburger menu
+  const notSmallScreen = useMediaQuery('(min-width: 768px)');
```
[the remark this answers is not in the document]
leekahung: That's a good idea.

I'll use display: none and toggle between that and display: initial depending on media query smallscreen in src/styles/layout.scss

I've produced the same effect without using the custom hook.

src/components/layout.tsx
```diff
@@ -1,52 +1,42 @@
...
-          '60px 1fr',
-        );
-      } else {
-        window.document.documentElement.style.setProperty(
-          '--layoutGridRows',
-          '100px 1fr',
-        );
-      }
-    }
+  if (notMedium) {
```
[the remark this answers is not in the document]
leekahung: Decided to use the same trick to solve for this custom hook and moving the stylings to src/styles/layout.scss.

.layout will remain the same, but two new classes .withfooter .noFooter is introduced and can be toggle depending on whether or not footer exists

onetrickwolf (approved):
Looks good not an expert on tailwind best practices but I don't see any big issues. Would probably be best to get this merged and address smaller issues in future PRs if there's anything left as this is pretty large 👍

FunctionDJ (approved):
LGTM

PR #89, opened 2023-04-26
Adding in rename-images script

Pauan:
You run the script like this:

```
python scripts/rename-images.py ../AWBW-Replay-Player/
```

And it will rename all of the tile files.

(The argument is the path to https://github.com/DeamonHunter/AWBW-Replay-Player)

### DIFF
--- a/scripts/rename-images.py
+++ b/scripts/rename-images.py
@@ -1,0 +1,23 @@
+import argparse
+import shutil
+import re
+from pathlib import Path
+
+parser = argparse.ArgumentParser()
+
+parser.add_argument("path")
+
+args = parser.parse_args()
+
+tile_dir = Path(args.path) / 'AWBWApp.Resources' / 'Textures' / 'Map' / 'AW2'
+
+exclude = ["Pipe", "River", "Road", "Sea", "Shoal"]
+
+for entry in tile_dir.iterdir():
+    if entry.is_dir():
+        co = entry.name
+
+        if co not in exclude:
+            for entry in entry.iterdir():
+                name = re.sub(r'^([^\-]+\-)', "\\1" + co + "-", entry.name.replace("_", "-")).lower()
+                entry.rename(entry.with_name(name))


megagreg72:
why / when would we run this?
are we expected to run this just once (like a data migration)?
does it download images from that other github project and convert it to our image-naming format (`-` vs `_`)?

megagreg72:
> why / when would we run this? are we expected to run this just once (like a data migration)? does it download images from that other github project and convert them to our image-naming format (`-` vs `_`)?

ideally, most of the common questions about the script like "why / when /how should I run this" should be answered by comments inside the script instead of just here on github.  You could even make the script check for a `--help` flag and show some usage info, if you are feeling fancy (but that might be overkill, and is definitely overkill if we only plan to run this once)

scripts/rename-images.py
```diff
@@ -0,0 +1,23 @@
...
+
+exclude = ["Pipe", "River", "Road", "Sea", "Shoal"]
+
+for entry in tile_dir.iterdir():
+    if entry.is_dir():
+        co = entry.name
+
+        if co not in exclude:
+            for entry in entry.iterdir():
+                name = re.sub(r'^([^\-]+\-)', "\\1" + co + "-", entry.name.replace("_", "-")).lower()
```
megagreg72, line 22: a lot of stuff happens on this line, consider breaking it into more lines and/or adding comments

Pauan:
It's a custom script requested by @FemboyDeveloper. It's only run once, to rename the image files from https://github.com/DeamonHunter/AWBW-Replay-Player.

megagreg72 (changes requested):
explain why / when / how to use this script with in-code comments.
optionally, clarify the one complicated line of code.

scripts/rename-images.py
```diff
@@ -0,0 +1,23 @@
+import argparse
+import shutil
+import re
+from pathlib import Path
+
+parser = argparse.ArgumentParser()
```
megagreg72, line 6:
```suggestion
from pathlib import Path

"""
This script was used to copy images from https://github.com/DeamonHunter/AWBW-Replay-Player
and rename them into our preferred filename format, for example replacing _ with -
We should never need to run this script again, but we are keeping it under source control just in case.
"""

parser = argparse.ArgumentParser()
```

megagreg72: ^ I paraphrased your explanation, feel free to edit the comment further

FunctionDJ (commented):
adding a python script to the repo doesn't make a lot of sense to me because of the additional setup involved.
i'll only approve if you or someone else provides a script that can be run with node or ts-node.

leekahung:
I think we can use something like fs to help Femboy process those files. That way we can incorporated it with Node on package.json (see example below).

https://user-images.githubusercontent.com/14917816/235296506-fd9fa159-25bd-43fd-a168-51abe84f093f.mov

FemboyDeveloper:
This was very useful! But as others have pointed out, it wouldnt be used in the repo/its only a one time use script so we don't really need it to be a file.

Pauan:
@FunctionDJ @leekahung There is absolutely no benefit to using Node in this case, and in fact it would make the script worse. This isn't a script that is run by users or developers, it's a one-time script run by @FemboyDeveloper only. And Python is the most popular language, so it's basically guaranteed that you will have Python already installed. Python is a fantastic language to create quick one-off scripts.

leekahung:
@Pauan I'm fine with either actually since it's just a one-off script. Python is certainly good for this type of work. Bash should work for this too. But I was thinking if do want to incorporate it down the line, we could also use fs. I think that's reasonable.

PR #93, opened 2023-04-30
Update NextAuth

MatthewKuKanich:
Rewrote the Next-Auth API/auth script to include a GitHub and Google OAuth sign-in provider, as well as added a passwordless email sign-in for backup access. I added a pages option with the sign-in and sign-out pages linked (will add those later). I also added a callbacks option, currently there is only the Sign-in callback which will be modified. I plan on using the callbacks to keep track of the session token to allow for account linking (Using Discord for sign-in, then converting to a regular User). I will do this with a JWT and session decryption callback.

### DIFF
--- a/src/pages/api/auth/[...nextauth].ts
+++ b/src/pages/api/auth/[...nextauth].ts
@@ -1,18 +1,65 @@
 import NextAuth from "next-auth";
-import { AppProviders } from "next-auth/providers";
 import DiscordProvider from "next-auth/providers/discord";
+import GithubProvider from "next-auth/providers/github";
+import GoogleProvider from "next-auth/providers/google";
+import EmailProvider from "next-auth/providers/email";
 
-const providers: AppProviders = [];
-const clientId = process.env.DISCORD_CLIENT_ID;
-const clientSecret = process.env.DISCORD_CLIENT_SECRET;
+const providers = [];
 
-if (clientId !== undefined && clientSecret !== undefined) {
+if (process.env.DISCORD_CLIENT_ID && process.env.DISCORD_CLIENT_SECRET) {
   providers.push(
     DiscordProvider({
-      clientId,
-      clientSecret,
-    }),
+      clientId: process.env.DISCORD_CLIENT_ID,
+      clientSecret: process.env.DISCORD_CLIENT_SECRET,
+    })
   );
 }
 
-export default NextAuth({ providers });
+if (process.env.EMAIL_SERVER && process.env.EMAIL_FROM) {
+  providers.push(
+    EmailProvider({
+      server: process.env.EMAIL_SERVER,
+      from: process.env.EMAIL_FROM,
+    })
+  );
+}
+
+if (process.env.GITHUB_CLIENT_ID && process.env.GITHUB_CLIENT_SECRET) {
+  providers.push(
+    GithubProvider({
+      clientId: process.env.GITHUB_CLIENT_ID,
+      clientSecret: process.env.GITHUB_CLIENT_SECRET,
+    })
+  );
+}
+
+if (process.env.GOOGLE_CLIENT_ID && process.env.GOOGLE_CLIENT_SECRET) {
+  providers.push(
+    GoogleProvider({
+      clientId: process.env.GOOGLE_CLIENT_ID,
+      clientSecret: process.env.GOOGLE_CLIENT_SECRET,
+    })
+  );
+}
+
+export default NextAuth({
+  providers,
+
+  // The Email provider requires a database, we can comment this provider out until that is setup
+  // database: process.env.DATABASE_URL
+  
+  secret: process.env.SECRET,
+
+  pages: {
+    signIn: "/auth/signin",
+    signOut: "/auth/signout"
+  },
+
+  callbacks: {
+    async signIn({ user, account, profile, email, credentials }) {
+      return true;
+    },
+  },
+
+  debug: false,
+});
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "clsx": "^1.1.1",
     "next": "^13.2.1",
     "next-auth": "^4.14.0",
+    "nodemailer": "^6.9.1",
     "pixijs": "^7.1.4",
     "prisma-json-types-generator": "^2.3.1",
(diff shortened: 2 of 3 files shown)


src/pages/api/auth/[...nextauth].ts
```diff
@@ -1,18 +1,65 @@
...
-import { AppProviders } from "next-auth/providers";
 import DiscordProvider from "next-auth/providers/discord";
+import GithubProvider from "next-auth/providers/github";
+import GoogleProvider from "next-auth/providers/google";
+import EmailProvider from "next-auth/providers/email";
 
-const providers: AppProviders = [];
-const clientId = process.env.DISCORD_CLIENT_ID;
-const clientSecret = process.env.DISCORD_CLIENT_SECRET;
+const providers = [];
```
Comment (FunctionDJ) [inline src/pages/api/auth/[...nextauth].ts:7]: doesn't removing the type make TS infer `providers` type to be `never[]`?

src/pages/api/auth/[...nextauth].ts
```diff
@@ -1,18 +1,65 @@
...
 
-if (clientId !== undefined && clientSecret !== undefined) {
+if (process.env.DISCORD_CLIENT_ID && process.env.DISCORD_CLIENT_SECRET) {
```
Comment (FunctionDJ) [inline src/pages/api/auth/[...nextauth].ts:9]: explicit checks like before (`!== undefined`) should be preferred over vague truthy checks

src/pages/api/auth/[...nextauth].ts
```diff
@@ -1,18 +1,65 @@
...
   providers.push(
     DiscordProvider({
-      clientId,
-      clientSecret,
-    }),
+      clientId: process.env.DISCORD_CLIENT_ID,
```
Comment (FunctionDJ) [inline src/pages/api/auth/[...nextauth].ts:12]: do the TS type gates work properly when you reference the env directly again like this instead of a variable like before?

src/pages/api/auth/[...nextauth].ts
```diff
@@ -1,18 +1,65 @@
...
+      clientSecret: process.env.DISCORD_CLIENT_SECRET,
+    })
   );
 }
 
-export default NextAuth({ providers });
+if (process.env.EMAIL_SERVER && process.env.EMAIL_FROM) {
```
Comment (FunctionDJ) [inline src/pages/api/auth/[...nextauth].ts:18]: is there anything that could break in development when providers that were enabled before are not enabled anymore? e.g. due to removing or changing the environment variables.
i was thinking that we could theoretically enforce specifying all environment variables, but then developers would need to specify a mail server too.

Comment (FunctionDJ) [inline src/pages/api/auth/[...nextauth].ts:18]: please update the .env example to document the new possible variables and their effect. e.g.:
```env
# Allows passwordless / email sign up and sign in. Optional.
# EMAIL_SERVER=
# EMAIL_FROM=
```

src/pages/api/auth/[...nextauth].ts
```diff
@@ -1,18 +1,65 @@
...
+    GithubProvider({
+      clientId: process.env.GITHUB_CLIENT_ID,
+      clientSecret: process.env.GITHUB_CLIENT_SECRET,
+    })
+  );
+}
+
+if (process.env.GOOGLE_CLIENT_ID && process.env.GOOGLE_CLIENT_SECRET) {
+  providers.push(
+    GoogleProvider({
```
Comment (FunctionDJ) [inline src/pages/api/auth/[...nextauth].ts:38]: does having google as a provider impact the users privacy any differently than having discord for example?

src/pages/api/auth/[...nextauth].ts
```diff
@@ -1,18 +1,65 @@
...
+      clientId: process.env.GOOGLE_CLIENT_ID,
+      clientSecret: process.env.GOOGLE_CLIENT_SECRET,
+    })
+  );
+}
+
+export default NextAuth({
+  providers,
+
+  // The Email provider requires a database, we can comment this provider out until that is setup
```
Comment (FunctionDJ) [inline src/pages/api/auth/[...nextauth].ts:48]: that would just be our own database, wouldn't it?

FunctionDJ (changes requested):

FunctionDJ:
ayo what, i deleted my branch and it closed this PR?
wasn't intentional.

FunctionDJ:
we can still recover this PR even if it's gonna be manual.

FemboyDeveloper:
smh function being toxic 😳

MatthewKuKanich:
All good, I'll update you

src/pages/api/auth/[...nextauth].ts
Comment (MatthewKuKanich) [reply src/pages/api/auth/[...nextauth].ts:12]: Yes, the environmental variables can be referenced directly. The type definitions can be set by both declaring them in code like before as well as in the type declaration file

src/pages/api/auth/[...nextauth].ts
Comment (MatthewKuKanich) [reply src/pages/api/auth/[...nextauth].ts:18]: I updated my code to set defaults. This should allow everything to function as normal if we decide to stop using or change the env variables.

Comment (MatthewKuKanich) [reply src/pages/api/auth/[...nextauth].ts:18]: Added that and matched it with the defaults

src/pages/api/auth/[...nextauth].ts
Comment (MatthewKuKanich) [reply src/pages/api/auth/[...nextauth].ts:38]: We should be able to provide the user a just as private login and access control functionality using google as an additional provider. We will have control on user data collection (and lack thereof in our case). We will come up with the necessary scopes to enable us to collect as little data as possible, See: https://developers.google.com/identity/protocols/oauth2/web-server#identify-access-scopes

src/pages/api/auth/[...nextauth].ts
Comment (MatthewKuKanich) [reply src/pages/api/auth/[...nextauth].ts:48]: Yes, I haven't fully looked into everything regarding our database, so I wasn't going to implement anything until I have a further understanding. I'd be happy to learn about it from you

src/pages/api/auth/[...nextauth].ts
Comment (MatthewKuKanich) [reply src/pages/api/auth/[...nextauth].ts:7]: No, since it's an empty array TS will infer ```providers``` type to be ```any[]```. Regardless, I added it back as a failsafe.

src/pages/api/auth/[...nextauth].ts
Comment (MatthewKuKanich) [reply src/pages/api/auth/[...nextauth].ts:9]: Readded and reworked checking with a set of default variables

PR #95, opened 2023-04-30
New Home Page

thatchaseb:
+ Home Page (in pages rendered at /home)
+ orangeStarCos.png (used in home.tsx)
+ react-multi-carousel dependantcy (used in home.tsx)


public/img/orangeStarCos.png
FemboyDeveloper, line 1: This image needs to be way smaller, it's almost 1MB, try to keep images at 100kb or less, 200 being the max for images we really need to be high definition. I recommend using https://squoosh.app/ to do so

src/pages/home.tsx
```diff
@@ -0,0 +1,349 @@
...
+      content: "";
+      z-index: -1;
+    }
+
+    .start-new-game-text {
+      margin: 30px 0px 0px 10px;
+      font-size: 20px;
+      font-weight: bold;
+    }
+  `
```
FemboyDeveloper, line 191: We are currently using [Tailwind CSS](https://v2.tailwindcss.com/docs) for our styling so you don't need to write out css like this. Also it's not best practice to keep css inside the same file, usually better to have your own home.css file.

src/pages/home.tsx
```diff
@@ -0,0 +1,349 @@
...
+  
+  return (
+    
+    <div style={{marginBottom:"0", color: "#333", fontFamily: "Arial, sans-serif"}}>
```
FemboyDeveloper, line 195: Don't use inline styles, it's a really bad practice and it can cause a lot of issues down the road. Instead use Tailwind classes, which is kinda like writing inline styles but in a cleaner/more maintainable manner. 

It make your code harder to read and understand, its usually best practice to separate the elements from the styling.

FemboyDeveloper (changes requested):
Left some requests, biggest issue is the way the styling was handled. I also see a lot of hardcoded values which worries me a bit (such as widht = 33.33% - 48px?). I'd try to avoid those if possible and use more responsive sizes such as vw or % which I see you have been using!

Let me know if you have any questions, keep up the good work!

src/pages/home.tsx
```diff
@@ -0,0 +1,349 @@
+import Carousel from 'react-multi-carousel';
+import React from 'react';
+import 'react-multi-carousel/lib/styles.css';
+
+export default function HomePage() {
+  
+  const responsive = {
```
leekahung, line 7: We do have existing responsive points with Tailwind, namely for tablet (min-width: 480px), which is similar to your mobile, smallscreen (min-width: 768px), which is somewhat similar to your tablet, and laptop (min-width: 1024px), which is similar to your desktop, all set within in tailwind.config.js. I think you can use some of those for styling. You can use the @screen rules in the CSS/SASS files for specific CSS classes.

For breakpoint that doesn't exist at the moment, I think you can add to it, like for superLargeDesktop. We can add a "largeDesktop" breakpoint which uses (min-width: 3000px) to tailwind.config.js

src/pages/home.tsx
leekahung: I agree with @FemboyDeveloper. We can move these stylings to Tailwind since we have things already set up for it or to a dedicated CSS/SASS file if you need to call specific classes.

leekahung (changes requested):
Adding some more feedback to src/pages/home.tsx

src/pages/home.tsx
thatchaseb: The resposive const is specifically needed for the react-multi-carousel dependancy

leekahung: Ah okay. I was not familiar with that library. If it's necessary for that functionality, I'm fine with keeping what it is for responsive.

public/img/orangeStarCos.png
FunctionDJ: large images in the repo are fine https://nextjs.org/docs/pages/api-reference/components/image

FemboyDeveloper:
@thatchaseb any updates? Should I close this PR for you?

PR #104, opened 2023-05-06
let tools detect node version

megagreg72:
tested locally with the tool [fnm](https://github.com/Schniz/fnm), which installed the correct version of node.


FunctionDJ (approved):
i'm a bit hesitant to add configuration files that are only needed by the minority of developers.
but i do see the point.
could this also be solved by just loosening up the required node version? the pnpm node version pin is only because of the starter template. i think we can get away with node 18 or higher in general.
otherwise, i'm fine with it.

megagreg72:
I fixed the build errors, but I still can't merge.

PR #105, opened 2023-05-07
fix prettier so VSCode extension is not required

megagreg72:
fix prettier scripts so VSCode extension is not required.
document how to lint.
also fixes some pre-existing linter and type issues so we get green checkmarks from CI/CD again.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,6 +4,9 @@ module.exports = {
     'plugin:@typescript-eslint/recommended',
     'plugin:react/recommended',
     'plugin:react-hooks/recommended',
+    // prettier and eslint configuration based on:
+    // https://blog.logrocket.com/using-prettier-eslint-automate-formatting-fixing-javascript/#remove-conflicting-rules-run-serially
+    'prettier', // Make sure this is last
   ],
   parserOptions: {
     ecmaVersion: 2018,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,9 +4,7 @@ module.exports = {
     'plugin:@typescript-eslint/recommended',
     'plugin:react/recommended',
     'plugin:react-hooks/recommended',
-    // prettier and eslint configuration based on:
-    // https://blog.logrocket.com/using-prettier-eslint-automate-formatting-fixing-javascript/#remove-conflicting-rules-run-serially
-    'prettier', // Make sure this is last
+    'plugin:prettier/recommended',
   ],
   parserOptions: {
     ecmaVersion: 2018,
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,3 +1,4 @@
 pnpm-lock.yaml
+node_modules/
 .next/
 dist/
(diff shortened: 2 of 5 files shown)


src/pages/match/index.tsx
```diff
@@ -4,7 +4,7 @@ import Link from 'next/link';
 import { trpc } from 'utils/trpc';
 
 const MatchRow = ({ match }: { match: Match }) => {
-  const { mapMetaData, mapTiles } = match.matchState;
+  const { mapMetaData, mapTiles } = match.matchState as any;
```
megagreg72, line 7: unrelated bug which was causing us to fail CI/CD checks

package.json
```diff
@@ -25,8 +25,12 @@
     "migrate": "prisma migrate deploy",
     "start": "cross-env NODE_ENV=production node dist/server/prodServer.js",
     "studio": "prisma studio",
-    "lint": "eslint --cache --ext \".js,.ts,.tsx\" --report-unused-disable-directives --report-unused-disable-directives src --max-warnings=10 && prettier --check .",
-    "lint-fix": "pnpm lint --fix",
+    "lint:eslint": "eslint src --ext \".js,.ts,.tsx\" --cache --report-unused-disable-directives --max-warnings=10",
+    "lint:eslint:fix": "pnpm lint:eslint --fix",
+    "lint:prettier": "prettier --check .",
```
FunctionDJ, line 30: isn't the point of adding prettier as an eslint plugin that eslint will also report prettier issues? (therefore removing the need to run prettier standalone as well)

megagreg72: I think the plugin just removes overlapping rules so they don't conflict with each other.  not sure, though

megagreg72: lint-fix was broken before and it works now when I test it (fixes eslint issues and prettier issues), but I'm not sure if this is the ideal setup.

megagreg72:
At @FunctionDJ 's suggestion, we now use the prettier plugin so it runs as part of eslint.  However, we also want to run prettier on some non-code files like README.md, so I decided to keep the `lint:prettier` scripts too.

.prettierignore
```diff
@@ -1,3 +1,4 @@
 pnpm-lock.yaml
+node_modules/
```
megagreg72, line 2: prettier actually ignores node_modules by default
https://stackoverflow.com/questions/53333181/how-to-ignore-node-modules-from-prettier
let me know if you prefer to exclude explicitly, or just rely on the default behavior.

.eslintrc.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
     'plugin:@typescript-eslint/recommended',
     'plugin:react/recommended',
     'plugin:react-hooks/recommended',
+    'plugin:prettier/recommended',
```
megagreg72, line 7: run prettier as part of eslint

package.json
megagreg72: I fixed our `.eslintrc` to use the plugin, thanks for the suggestion.  I am keeping the prettier scripts so we can lint non-code files.

FemboyDeveloper:
@megagreg72 is this ready for merging?

megagreg72:
yup, ready to merge :+1:

FemboyDeveloper (approved):
LGTM!

PR #106, opened 2023-05-07
add lots of comments about tiles to functions-branch

megagreg72:


src/shared/match-logic/tiles.ts
```diff
@@ -76,84 +82,200 @@ const pipeTileProperties: TileProperties = {
...
+   * which is about 1/10 actual health.
+   *
+   * However, *we* are currently planning to store health as a number from 0 to 1,
+   * and display it as decimal from 1 to 10 *exactly* like "8.7".
+   *
+   * So instead of dividing by 10 (with rounding), we multiply by 10.
+   */
+  const visualHp = hp * 10;
+  return tileProperties[tileType].defenseStars * visualHp;
+};
```
megagreg72, line 256: I *think* I figured out what this function does. Let me know if this is correct :)

src/shared/match-logic/tiles.ts
```diff
@@ -76,84 +82,200 @@ const pipeTileProperties: TileProperties = {
...
+ * a unit on that tile takes from attacks.
+ */
 export const getTerrainDefenseStars = (tileType: TileType) =>
-  movementCostMap[tileType].defenseStars;
+  tileProperties[tileType].defenseStars;
 
-export const getUnitTerrainDefense = (hp: number, tileType: TileType) =>
-  movementCostMap[tileType].defenseStars * 10 * hp;
+/**
+ * Caculate the defense of a unit.
```
megagreg72, line 231:
```suggestion
 * Calculate the defense of a unit.
```

src/shared/match-logic/tiles.ts
```diff
@@ -2,66 +2,72 @@ import { TileType } from "server/schemas/tile";
...
+ * Specifically, if a map tile is passable for infantry,
+ * we consider it to be a "land" tile.
+ */
+const landMovementCosts = {
+  boots: 1,
+  air: 1,
+  pipe: null,
+  sea: null,
+  lander: null,
+  // `as const satisfies ...` allows us to essentially define a custom type with exactly
```
FunctionDJ, line 38: `as const` is not needed for this. all it does is make all properties `readonly` and turns the value types from regular primitives to literals.
i'm using `satisfies` in many places so for consistency you'd either need to copy this explanation to every instance or you just don't and let someone use Google or watch a TS tutorial for a change.
there are many other TS concepts i consider worth explaining like mapped types because unless you know what they're named, there's no keyword you can put into a search engine to learn about it.

src/shared/match-logic/tiles.ts
```diff
@@ -76,84 +82,200 @@ const pipeTileProperties: TileProperties = {
...
+ * @param tileType the tile which the unit is on, e.g. "plains"
+ */
+export const getUnitTerrainDefense = (hp: number, tileType: TileType) => {
+  /**
+   * Historically, this game series has *stored* health as a number from 1 to 100,
+   * and *displayed* health as a number from 1 to 10, by dividing by 10 with rounding.
+   * The old damage formula is based on "visual health"
+   * which is about 1/10 actual health.
+   *
+   * However, *we* are currently planning to store health as a number from 0 to 1,
```
FunctionDJ, line 249: this `*` was a mistake. we store HP as 0-100 https://github.com/WarsWorld/WarsWorld/blob/cc52eb92d3c5835444d5ea59feab544c7e932a46/src/server/schemas/unit-traits.ts#L6

FunctionDJ (commented):
the code changes make the code more complicated in my opinion. "manMadeMovementCosts" is a new concept that needed to be explained instead of just leaving the previous system that simply created some reusable tile properties.
there's also more redundancy for lab, comtower, unusedSilo and usedSilo.
the file is also now contains about 35 more lines of code that don't change the behaviour of the code as far as i can tell and which don't make the code more maintainable or understandable as i've explained at the start.

src/shared/match-logic/tiles.ts
megagreg72: ah, you are right.  I used it with `as const` out of habit but in this case that isn't needed

src/shared/match-logic/tiles.ts
```diff
@@ -2,66 +2,72 @@ import { TileType } from "server/schemas/tile";
...
-    lander: null,
-  },
-};
+/**
+ * Movement costs which are shared by all ground-based map tiles
+ * like forests, rivers and cities.
+ * Specifically, if a map tile is passable for infantry,
+ * we consider it to be a "land" tile.
+ */
+const landMovementCosts = {
```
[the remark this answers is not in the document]
megagreg72: I was thinking that "Movement costs which are shared" and "satisfies Partial<TileMovementCosts>;" would clarify it, but I agree that `commonLandMovementCosts` is better.

src/shared/match-logic/tiles.ts
```diff
@@ -76,84 +74,184 @@ const pipeTileProperties: TileProperties = {
...
+   * The damage formula is based this "visual health".
+   */
+  const visualHp = hp / 10;
+  return tileProperties[tileType].defenseStars * visualHp;
+};
 
+/**
+ * Every turn, units get a certain number of movement points
+ * which they can spend by moving.
+ * Every unit has exactly one "movement type", for example tanks have type "treads".
```
FunctionDJ, line 237: i think the info of this specific line is common knowledge in the AW community.

src/shared/match-logic/tiles.ts
```diff
@@ -76,84 +74,184 @@ const pipeTileProperties: TileProperties = {
...
+ *
+ * @param tileType The tile which the unit is trying to enter, e.g. "plains"
+ * @param movementType The movement type of the unit, e.g. "treads"
+ * @param weather The current weather, e.g. "rain"
```
FunctionDJ, line 241: the examples for type values are not needed because a developer can just use their editor/IDE to see which values are possible or use jump-to-definition.

FunctionDJ (changes requested):

src/shared/match-logic/tiles.ts
```diff
@@ -76,84 +74,184 @@ const pipeTileProperties: TileProperties = {
...
+ *
+ * Note: Movement costs vary based on weather,
+ * so we need to either:
+ * 1) create one of these objects for each possible weather
+ * 2) create one of these objects for "normal" weather and
+ * perform some final weather-based modifications in the getter function.
+ *
+ * Currently we use approach (2).
+ *
+ * TODO: Create a TileType for a "broken pipe seam".
```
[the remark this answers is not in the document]
megagreg72: I think it is weird that broken pipe seams are considered a variant of plains, but used missile silo is not a variant of missile silo.  Both variants have gameplay implications (Jake's power for plains).
But thanks for letting me know how it is currently handled 👍

src/shared/match-logic/tiles.ts
```diff
@@ -76,84 +74,184 @@ const pipeTileProperties: TileProperties = {
...
+ * to *enter* each type of tile, for each "movement type".
+ * `null` means impassible terrain.
+ *
+ * Note: Movement costs vary based on weather,
+ * so we need to either:
+ * 1) create one of these objects for each possible weather
+ * 2) create one of these objects for "normal" weather and
+ * perform some final weather-based modifications in the getter function.
+ *
+ * Currently we use approach (2).
```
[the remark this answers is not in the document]
megagreg72: I was thinking we might want to pre-calculate the movement costs as space-vs-time optimization, but we can probably get the best of both worlds by just memoizing the getter function.

src/shared/match-logic/tiles.ts
```diff
@@ -2,66 +2,64 @@ import { TileType } from "server/schemas/tile";
...
-    air: 1,
-    pipe: null,
-    sea: null,
-    lander: null,
-  },
+/**
+ * All man-made structures (roads, buildings) have the same movement costs,
+ * with two exceptions:
+ * - Pipes are impassible except by piperunners.
+ * - Any building which *could have* produced a unit
```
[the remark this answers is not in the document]
megagreg72: I can see where you are coming from, but I prefer this.
I like that it provides a logical *explanation* for why piperunners can move through bases etc.
I suppose it is inevitable that everyone likes their own code 😄

src/shared/match-logic/tiles.ts
```diff
@@ -2,66 +2,64 @@ import { TileType } from "server/schemas/tile";
...
-    sea: 1,
-    lander: 1,
-  },
+const buildingTileProperties: TileProperties = {
+  /** All buildings provide 3 defense, except for the HQ which provides 4. */
+  defenseStars: 3,
+  movementCosts: manMadeMovementCosts,
 };
 
+/** Pipes and (unbroken) pipe seams are exactly the same */
```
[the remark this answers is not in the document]
megagreg72: it saves a dev 5 seconds of scrolling down to see usage 🤷‍♂️

src/shared/match-logic/tiles.ts
```diff
@@ -76,84 +74,184 @@ const pipeTileProperties: TileProperties = {
...
+ * 1) create one of these objects for each possible weather
+ * 2) create one of these objects for "normal" weather and
+ * perform some final weather-based modifications in the getter function.
+ *
+ * Currently we use approach (2).
+ *
+ * TODO: Create a TileType for a "broken pipe seam".
+ *
+ * Also we *could* combine usedSilo and unusedSilo
+ * because they have the same defense and movement properties.
```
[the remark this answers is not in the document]
megagreg72: ok, I removed the kinda-suggestion and just made two notes on `TileType` itself about pipe seam => plains and silo => used silo

src/shared/match-logic/tiles.ts
megagreg72: even with AW knowledge, "movement type" could be something like `'normal' | 'transport-unload'` or `'normal' | 'lash-power' | 'sturm'`

src/shared/match-logic/tiles.ts
megagreg72: fair enough, I don't mind removing these examples because I explained movement type above.

megagreg72: done 👍

src/shared/match-logic/tiles.ts
FunctionDJ: https://awbw.fandom.com/wiki/Units#Movement

FunctionDJ (changes requested):

src/shared/match-logic/tiles.ts
```diff
@@ -76,84 +74,172 @@ const pipeTileProperties: TileProperties = {
...
+};
 
+/**
+ * Every turn, units get a certain number of movement points
+ * which they can spend by moving.
+ * Every unit has exactly one "movement type", for example tanks have type "treads".
+ *
+ * @param tileType The tile which the unit is trying to enter
+ * @param movementType The movement type of the unit
+ * @param weather The current weather
```
FunctionDJ, line 229: do the movementType and weather `@param`s really add substantial information here? i think they can be omitted.

megagreg72: I will not remove these. I added back examples for `tileType` and `movementType` to make them more useful again.

src/shared/match-logic/tiles.ts
```diff
@@ -76,84 +74,173 @@ const pipeTileProperties: TileProperties = {
...
+ */
+export const getUnitTerrainDefense = (hp: number, tileType: TileType) => {
+  /**
+   * We *store* health as a number from 1 to 100,
+   * and *display* health as a number from 1 to 10, by dividing by 10 with rounding.
+   * The damage formula is based this "visual health".
+   */
+  const visualHp = hp / 10;
+  return tileProperties[tileType].defenseStars * visualHp;
+};
```
megagreg72, line 220: fixed hp bug

megagreg72:
I we resolved all of Function's suggestions except for some comment-related suggestions where we will "agree to disagree".
I'm ready to merge this whenever.

PR #113, opened 2023-05-10
Fixing typo in Navbar and updates to Match page

leekahung:
Several things are made in this PR. Firstly, there's the minor typo fixed by correcting NavLinkMatches to:

```JSX
<NavItem text="Your Games" location="/your-matches" />
<NavItem text="Current Games" location="/your-matches#currentGames" />
<NavItem text="Completed Games" location="/your-matches#completedGames" />
```

It was previously:

```JSX
<NavItem text="Current Games" location="/your-matches" />
<NavItem text="Current Games" location="/your-matches#currentGames" />
<NavItem text="Completed Games" location="/your-matches#completedGames" />
```

Second, I've attempted to convert our previous Match page with PixiMatch, which does appear as an invisible canvas (the element is rendered, but I can't see). I'm not sure if it's because the assets are not yet build in or if it require the Pixi branch to be finished. In either case, I've temporarily turned off their height and width. Will likely have this reverted back on once we have working assets.

However, I did manage to put the PlayerBoxes back into the Match page. (see below)

Screen Shot 2023-05-11 at 4 22 45

I've refactored PlayerBox into its own component and using it inside main.tsx where the match information is located with PixiMatch.

If playerInMatch is not null/undefined, it'll populate the same as before with the proper CO, Army Icon, etc. To that end, I've tweaked the typing for PlayerInMatch to include the type Army found in our existing schemas.

If playerInMatch is null/undefined (which would probably occur if the other player slot is not filled), it'll show a white gradient with placeholders.

Lastly, I've noticed global.scss has been duplicated (one placed inside /src/frontend/styles/ and one placed inside /src/frontend/styles/grid/). I've used vimdiff to verify this and deleted the one not under grid.


leekahung:
Updating initial PR post to include recent changes in commits.

PR #112, opened 2023-05-10
Removed pnpm and tidied up package.json scripts

FunctionDJ:

PR #119, opened 2023-05-14
remove postinstall

megagreg72:
https://github.com/WarsWorld/WarsWorld/pull/112#discussion_r1190472950
prisma runs this on its own.  tested with a clean install


FunctionDJ (approved):
I'm curious why the starter we used has this
https://github.com/trpc/examples-next-prisma-websockets-starter/blob/420be5a7916f4d91ae2307f484ce38c1f5c6c67c/package.json#L33
I've opened a thread on the tRPC Discord, maybe they'll remove it or it covers an edge case we don't know about. Either way, you tested it for our use case so it's fine.

PR #121, opened 2023-05-14
fix vscode auto-import for .d.ts files

megagreg72:
vscode will not suggest imports for files which are not included by tsconfig

we have some src files like `server-match-state.d.ts` where auto-import therefore doesnt work atm

alternatively, we could rename those files to be `.ts`


FunctionDJ:
using `.d.ts` files was a mistake because of `skipLibCheck`:
https://github.com/WarsWorld/WarsWorld/blob/6d67316ffe5f128b1a25467fcd2a36c0bd8938fb/tsconfig.json#LL6C21-L6C21
which makes TS not check `.d.ts` files. so in my local branch and moving forward, type files are only `.ts`.

megagreg72:
ok, I'll convert this to a draft and someone can make a PR which converts those `.d.ts` files.  im travelling atm so i cant do it easily, but i could do it later

megagreg72:
actually, it looks like `skipLibCheck` only skips `.d.ts` from _libraries_, not our own source code.  This PR worked fine for me locally, fixing types for our own `.d.ts` files.
https://www.typescriptlang.org/tsconfig#skipLibCheck

FemboyDeveloper (approved):
LGTM, just one line of code can't break the project... right? lol

PR #122, opened 2023-05-14
fix lint scripts, add line-length limit for comments

megagreg72:
The lint scripts were broken, so it no longer ran prettier for non-code files like `*.yaml` and `*.md`.
It looks like we already are respecting an 80-character limit for code (default settings?) but are not checking comment lengths.
I manually fixed a bunch of comments because it is not auto-fixable.

In my opinion, the point of an 80-character limit is to make it easier to view the "previous" and "current" versions of a file side-be-side in VSCode.  And one of the things you are going to want to read in that context is comments.


.github/dependabot.yml
```diff
@@ -1,7 +1,7 @@
 version: 2
 updates:
   - package-ecosystem: npm
-    directory: '/'
+    directory: "/"
```
megagreg72, line 4: I'm not sure familiar with `.yml` files, but a quick google search suggests that `"` is the same as `'` so our prettier rule could apply here too

package.json
```diff
@@ -18,15 +18,17 @@
...
-
     "dev": "npm run prisma:push && concurrently \"npm:dev:*\"",
-    "lint": "eslint src --ext \".js,.ts,.tsx\" --cache --report-unused-disable-directives --max-warnings=10",
-    "lint:fix": "npm run lint --fix",
+    "lint:eslint": "eslint src --ext \".js,.ts,.tsx\" --cache --report-unused-disable-directives --max-warnings=10",
+    "lint:eslint:fix": "npm run lint:eslint --fix",
+    "lint:prettier": "prettier --check .",
+    "lint:prettier:fix": "prettier --write .",
+    "lint": "npm run lint:prettier && npm run lint:eslint",
+    "lint:fix": "npm run lint:prettier:fix && npm run lint:eslint:fix",
```
megagreg72, line 27: fix lint scripts

CONTRIBUTING.md
```diff
@@ -21,12 +21,12 @@ In either case, a good branch name would be descriptive of what you plan to work
 Setup instructions can be found in our README.md!
 
 ## Linting
+
 Our CI/CD process will use a "linter" to check if your code is formatted in a readable way. However, it is up to you to actually fix any formatting issues.
 You can check for any problems by running `npm run lint`, and you can fix those problems manually or by running `npm run lint:fix`.
 Alternatively, if you are using VSCode, you should install the extension `rvest.vs-code-prettier-eslint`
 so that your code is automatically formatted every time you save a file.
 
-
```
megagreg72, line 29: ironic

src/shared/types/events.ts
```diff
@@ -28,8 +29,20 @@ export interface UnloadEvent extends UnloadAction {
 }
 
 export interface AttackEvent extends AttackAction {
-  defenderHP: number; // we could probably derive these as well if we just submit the rolled luck value
-  attackerHP?: number; // missing means no counter-attack
+  /**
+   * The new defender HP after the attack.
+   * TODO: Consider sending just the luck roll(s) for the event,
+   * and calculating HP later.
```
megagreg72, line 35:
```suggestion
   * TODO: Consider sending just the luck roll(s) for the event,
   *       and calculating HP later.
```

.eslintrc.js
```diff
@@ -24,6 +24,17 @@ module.exports = {
...
     "react/prop-types": "off",
     "@typescript-eslint/no-explicit-any": "off",
+    "max-len": [
+      "error",
+      {
+        code: 80,
+        tabWidth: 2,
+        ignoreComments: false,
+        ignoreUrls: true,
+        ignoreStrings: true,
```
FunctionDJ, line 34: if we don't ignore comments, i say we shouldn't ignore strings and template literals too. URLs sounds like it would be too annoying though and limiting them isn't all that useful for readability.

package.json
FunctionDJ: why so many scripts? our eslint config is using the prettier plugin. shouldn't there be no reason to run prettier standalone?
if eslint doesn't run on non-code files, can we fix that?

FunctionDJ (changes requested):

.eslintrc.js
megagreg72: right, I'm thinking about enabling that later, I just wanted to keep this PR small 👍

package.json
megagreg72: I think that eslint might be ignoring non-code files because of this
"eslint src --ext \".js,.ts,.tsx\"
```
or maybe eslint just ignores non-code files by default?  can't remember if I tested without that

megagreg72:
responded to comments

megagreg72:
https://github.com/WarsWorld/WarsWorld/pull/122#discussion_r1193188050
I believe that `eslint` is simply not designed to lint anything except for code files.
`prettier` is designed to lint pretty much anything, so I think we have to run `prettier` by itself if we want to lint non-code files like `README.md`.

megagreg72:
ha, our build checks prevented this PR from being merged because the linter was complaining about a comment which was too long.  That was definitely an intentional test of our build system 🤣

megagreg72:
hmm, actually it looks like linting takes about 6 seconds longer now (eyeballing some build times for the `lint` step in this PR compared to some other PRs)

megagreg72:
I saved about 3 seconds of build time by limiting which files `prettier` checks, to avoid double-checking the files `eslint` would check.
but at what cost? (my sanity)

PR #123, opened 2023-05-14
type overhaul

megagreg72:
It seems like `zod` is introducing code duplication atm, so we can either remove it and validate ourselves by writing functions like this, or just use `zod` better.
This is how I typically define sets of strings (essentially enums) without a library like `zod`.
I'll give it a try to see if it helps.


src/shared/types/sets/movement.ts
```diff
@@ -0,0 +1,18 @@
...
+  "pipe",
+] as const;
+
+export type MovementType = (typeof movementTypes)[number];
+
+export function isMovementType(x: any): x is MovementType {
+  return (
+    typeof x === "string" && (movementTypes as readonly string[]).includes(x)
+  );
+}
```
megagreg72, line 18: we might not even need zod if we use functions like this for runtime validation
but idk, I don't fully understand zod yet

src/shared/types/sets/movement.ts
```diff
@@ -0,0 +1,18 @@
...
+  "tires",
+  "air",
+  "sea",
+  "lander",
+  "pipe",
+] as const;
+
+export type MovementType = (typeof movementTypes)[number];
+
+export function isMovementType(x: any): x is MovementType {
```
FunctionDJ, line 14: we never need to validate a movement type. it's not a user input.

FunctionDJ (commented):
the purpose of zod is to create types and validation with a single structure: a schema.
if we remove zod, we need more code: some for the type, and then some for the validation. and they won't be kept in sync by default.

megagreg72:
after messing around a bit more, I think `zod` is useful but we could use it a bit better. I'll close this PR

PR #124, opened 2023-05-14
Change from pnpm to npm in deployment script

igoyak:
This should theoretically work, but doesn't actually in practice for a few reasons.


igoyak:
(It doesn't work anymore since the added dependency on pixijs, since pixijs depends on gl, and there's no prebuilt version of gl for Amazon linux, so to build it, python is required as a dependency, but this is not enough, because it also requires x11 dev libraries to compile)

It highlights that we need to go a different direction for deployment, this simple MVP solution of building on startup doesn't seem to work reliably. 
So we can merge this to stay more or less in sync with the project, or we can remove the deployment scripts all together until we decide on a more proper way forward.

FunctionDJ (approved):

PR #134, opened 2023-05-22
Add nodejs environment setup within docker compose

muhammadbna-dev:
### Current situation
- We have postgres maintained with docker-compose
- However, our node environment is maintained locally. The issue here is that all the contributors have to ensure that they have the right nodejs version `node:18.15.0.`. This inconsistency may cause issues down the line when differences in nodejs version amongst the contributors might affect the app functionality if the contributors did not maintain the right node version when running the codebase in their local environments.

### Suggestion
- Dockerize our nodejs environment. This ensures environment consistency. All contributors will now work `node:18.15.0` as defined in the Dockerfile

### What will change with this?
- Contributors will use a single command `docker-compose up` instead of a combination of commands: `npm run dev` and `docker-compose up`
- `DATABASE_URL` is now maintained in `docker-compose.yaml` instead of `.env`
- If there is any package changes in `package.json`, contributors will need to execute `docker-compose build` instead of `npm install`
- Change the implementation of seeding data in DB: If data already exits, don't need to seed. `prisma/seed.ts`
- Now that node environment is within a docker container, if we want to execute node commands we have to enter the container and execute the commands there: `docker-compose exec <service_name> bash` which enters the container and then in the container you can execute the commands.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM node:18.15
+COPY package-lock.json /app/package-lock.json
+COPY package.json /app/package.json
+WORKDIR /app
+RUN npm install
+COPY . /app/
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -10,5 +10,23 @@ services:
     environment:
       POSTGRES_PASSWORD: ${PGPASSWORD}
       POSTGRES_HOST_AUTH_METHOD: trust
+
+  app:
+    build:
+      dockerfile: Dockerfile
+    volumes:
+      - .:/app/
+      - /app/node_modules
+    ports:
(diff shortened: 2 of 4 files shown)


megagreg72:
current solution is for devs to run `nvm use` to switch to the correct node version.
still, I'm not against dockerizing the app.
as part of this PR, could you please update `README.md` and the `.env` files?

megagreg72:
one downside is that dockerizing might slow down local build time

FunctionDJ:
this change adds unnecessary setup complexity without a strong upside. managing nodejs versions isn't too hard, there's `nvm` and many others.

PR #141, opened 2023-05-30
Zod integers

megagreg72:
A Typescript "number" is an integer or float.  In most cases, we want integers only.


FemboyDeveloper (approved):
LGMT! Pretty sure we don't want floats to sneak in where we want ints, dont want no 1.5 inf loaded in a transport lol

PR #145, opened 2023-06-03
General code optimizations and clean-up

leekahung:
This PR is mainly bookkeeping stuff with small refactors to components and pages, along with some style updates.

Had also removed unused CSS classes from stylesheets, but repurposed homeBanner to run Tailwind Media Query for .cardsContainer.

Had also removed parts we're likely not going to revisit like `updateDocumentHeightCSSProperty` or `<meta>` tags outside of _app.jsx unless necessary. Also refactored `<Head>` components from components to Pages, where they'll be the most relevant with `<Head>` and `<title>`.

### DIFF
--- a/src/frontend/components/Footer.tsx
+++ b/src/frontend/components/Footer.tsx
@@ -1,38 +1,48 @@
 import Link from "next/link";
 
-export function Footer() {
-  const rLogoPath = "/img/layout/Reddit.png";
-  const dLogoPath = "/img/layout/Discord.png";
-  const gLogoPath = "/img/layout/GitHub.png";
+const rLogoPath = "/img/layout/Reddit.png";
+const dLogoPath = "/img/layout/Discord.png";
+const gLogoPath = "/img/layout/GitHub.png";
 
+const footerLinks1 = [
+  { text: "About us", href: "/about" },
+  { text: "Terms of Use", href: "/terms" },
+  { text: "Donations", href: "/donations" },
+];
+
+const footerLinks2 = [
+  { imgSrc: rLogoPath, imgAlt: "Reddit Logo", href: "/" },
+  { imgSrc: dLogoPath, imgAlt: "Discord Logo", href: "/" },
+  {
+    imgSrc: gLogoPath,
+    imgAlt: "GitHub Logo",
+    href: "https://github.com/warsWorld/WarsWorld/",
+  },
+];
+
+export function Footer() {
   return (
-    <footer className="@max-h-[200px] @w-full @flex @flex-col @items-center @justify-center @gap-4 @bg-gradient-to-t @from-black">
+    <footer className="@max-h-[200px] @w-full @flex @flex-col @items-center @justify-center @gap-4 @bg-gradient-to-t @from-black @pb-5">
       <nav className="@flex @gap-8">
-        <Link className="@text-base-a" href="/about">
-          About us
-        </Link>
-        <Link className="@text-base-a" href="/terms">
-          Terms of Use
-        </Link>
-        <Link className="@text-base-a" href="/donations">
-          Donations
-        </Link>
+        {footerLinks1.map((item) => (
+          <Link className="@text-base-a" key={item.text} href={item.href}>
+            {item.text}
+          </Link>
+        ))}
       </nav>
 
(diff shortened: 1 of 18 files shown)


src/frontend/components/Footer.tsx
```diff
@@ -1,38 +1,48 @@
 import Link from "next/link";
 
-export function Footer() {
-  const rLogoPath = "/img/layout/Reddit.png";
-  const dLogoPath = "/img/layout/Discord.png";
-  const gLogoPath = "/img/layout/GitHub.png";
+const rLogoPath = "/img/layout/Reddit.png";
+const dLogoPath = "/img/layout/Discord.png";
+const gLogoPath = "/img/layout/GitHub.png";
```
leekahung, line 5: Note that there's no existing assets at the moment at these paths. Will need to come back in the future

src/frontend/components/Home.tsx
```diff
@@ -4,49 +4,62 @@ import SmallContainer from "./Home/SmallContainer";
...
+        button={[
+          {
+            text: "Play Now",
+            link: "/register",
+          },
+          {
+            text: "Learn to Play",
+            link: "/learn",
+          },
+        ]}
```
leekahung, line 62: Should refactor this out to call the array rather than having the array inside. That way it'll be easy to update in the future if we're going to expand on this.

src/pages/match/[matchId].tsx
```diff
@@ -1,17 +1,10 @@
...
-    <>
-      <Head>
-        <meta name="viewport" content="width=device-width, initial-scale=1" />
-      </Head>
-
-      <Layout>
-        <PixiMatch />
-      </Layout>
-    </>
+    <Layout>
```
Comment (leekahung) [inline src/pages/match/[matchId].tsx:6]: Could try adding a Head here too for the Match Page

src/pages/match/[matchId].tsx
```diff
@@ -1,17 +1,10 @@
-import Head from "next/head";
 import { PixiMatch } from "frontend/components/match/main";
 import { Layout } from "frontend/components/layout";
 
 export default function Match() {
   return (
-    <>
-      <Head>
```
Comment (leekahung) [inline src/pages/match/[matchId].tsx:8]: You should actually keep the Head component. Could be useful for adding Page Title for the website.

leekahung (commented):
See notes.

src/frontend/components/Home.tsx
```diff
@@ -4,49 +4,62 @@ import SmallContainer from "./Home/SmallContainer";
...
+        title="There is a Space for You"
+        text="Whether you want to be hardcore or play fun crazy maps"
+        button={[
+          {
+            text: "Play Now",
+            link: "/register",
+          },
+          {
+            text: "Learn to Play",
+            link: "/learn",
```
leekahung, line 60: We do have a How to Play dummy page. These are related so we can link these to the same page.

src/frontend/components/Home.tsx
leekahung: Moved this outside of component.

src/pages/match/[matchId].tsx
Comment (leekahung) [reply src/pages/match/[matchId].tsx]: Included both Head and title elements.

src/pages/match/[matchId].tsx
Comment (leekahung) [reply src/pages/match/[matchId].tsx:8]: Done.

src/frontend/components/Home.tsx
leekahung: Changed "/learn" to "/howtoplay" for now. Could consider renaming route for that in the future to "learn". Much more straightforward.

FemboyDeveloper:
@leekahung did a small tweak to your branch.

Overall, love the cleanup, specially the small container map, makes the Home page file look much cleaner and readable. I did do some very minor tweaks, I re-added the buttons to the ThreeLinesText component because I think its a bit better that way since its pretty small.

I did once more delete homeBanner.scss BUT I did re-implement your laptop fix back simply by typing the classname "laptop:@flex-row" which is pretty cool that tailwind allows for quick media queries just like that.

Overall nice job! Love how the buttons are responsive now!

FemboyDeveloper (approved):
LGTM!

PR #146, opened 2023-06-03
Use Tailwind's themes over sass variables, tweak background

FemboyDeveloper:
Short PR, moved the variables.sass to tailwind.config.js so we can use the variables as themes, deleted old background pictures and implemented a new one closer to the one used in our Figma design.


leekahung (approved):
The theme switch from SASS to Tailwind looks alright, I'll approve.

PR #147, opened 2023-06-04
Reorganized files and folders for better clarity

FemboyDeveloper:
Reorganized the pages folder to be more streamlined, only pages that have dynamic content (such as match/1 or news/patch-1.1) will need a folder. Other pages can go by themelves without a folder and next.js will display them correctly.

## Home Page
the home page now has a BasicHome and a LoggedHome file, this is because when the user is logged in, the home page will change to become a player hub of sorts since logged in players wont visit the homepage since they already know the site.

## Sass
Renamed some files and tidied up the naming conventions used. Also deleted unused code.

## Layout
Also deleted the layout component from pages and made it a default on every page. This is because every single page will always have the navbar and footer. Currently I cant think of a page that merits not having either of these.


leekahung:
> ## Layout
> Also deleted the layout component from pages and made it a default on every page. This is because every single page will always have the navbar and footer. Currently I cant think of a page that merits not having either of these.

Kind of. I think it'll depend on what we're going to end up with as a footer. If it's just the copyright statement, I think it'll be fine to have it included in all pages.

If it's fully flushed out with the navigation component on top of it like what we have in the Figma, I think that footer would still be fine for pages like Home, How to Play, etc. But for pages like a Match Page, for example, since people wouldn't need to be scrolling there, it could be a bit excessive with both Navbar and Footer navigation present. It'll take up some of the space we want for the main game board (map, player boxes, etc.). Having a condition to remove it (or at least for the navigation portion of the footer) when needed is a good idea.

leekahung:
Not really sure what's causing the build to fail though.

FemboyDeveloper:
> 

Im fine with switching our footer up, but I think it should be included on every page. We always want that copyright notice and whatever other stuff we want to show

PR #150, opened 2023-06-06
News Page Updates

leekahung:
Created new components for News Page, including cards for news posts and a dedicated "featured" post at the top. Routed news to /news.

Included new uuid library for generating unique keys for news cards for now. (Would likely replaced in the future when we can mark each post with unique ids server-wise).

Update: Combining the two new card components (ArticleLinkCard and NewsCard) into one called LinkCard. LinkCard incorporates styling from both components.

Screen Shot 2023-06-07 at 13 12 25 (2)

Screen Shot 2023-06-07 at 13 12 33 (2)

Also created a new interface for cardInfo called ICardInfo inside LinkCard. Responsiveness is the same as before just with the updated cards:

https://github.com/WarsWorld/WarsWorld/assets/14917816/2439d78b-423c-450e-b8ac-4432e8482895


FemboyDeveloper (changes requested):
Overall looks good to me. 

Only issues I found are that the big/center card isnt very responsive on bigger sizes, everything else looks great!
image

leekahung:
> Only issues I found are that the big/center card isnt very responsive on bigger sizes, everything else looks great! image

Got it updated. Should be manageable for larger screens now. Used width 60vw for the feature picture and maxWidth 20vw for the textbox

leekahung:
Included updates to combine the two card components into one and removed the hover shadow.

src/frontend/components/news/FeaturedNews.tsx
```diff
@@ -0,0 +1,34 @@
...
+            and fair gameplay to Advance Wars. With careful adjustments made to
+            unit abilities and map layouts, commanders will find themselves
+            faced with more challenging decisions and a more balanced playing
+            field.
+          </p>
+        </div>
+      </div>
+    </div>
+  );
+}
```
FemboyDeveloper, line 34: The featured news should be a component/be able to take information from somewhere else rather than have it hardcoded, pretty minor thing.

src/pages/news/index.tsx
```diff
@@ -0,0 +1,87 @@
+import { FeaturedNews } from "frontend/components/news/FeaturedNews";
+import LinkCard from "frontend/components/layout/LinkCard";
+import Head from "next/head";
+import { v4 as uuidv4 } from "uuid";
```
FemboyDeveloper, line 4: are we using uuids here?

FemboyDeveloper (changes requested):
Very minor changes requested, overall looks great! Good job neo!

src/frontend/components/news/FeaturedNews.tsx
leekahung: > The featured news should be a component/be able to take information from somewhere else rather than have it hardcoded, pretty minor thing.

The hard-coded is just temporary, so it should be okay for the time being.

src/pages/news/index.tsx
leekahung: As mentioned in the OP, the uuid is a temporary thing that would be replaced  once we're able to generate the keys from the backend.

src/frontend/components/news/FeaturedNews.tsx
FemboyDeveloper: yeah I can accept that, we can change it later once we have some "realer" data lol

PR #151, opened 2023-06-06
Move files around, rename files from howtoplay emphasis to article

JaviTrek:
Small movement of files here and there, since these article cards and sections might be re-used in other pagews, its better for them to be known as article cards/sections.


CliveGlitch (commented):
Not so much to say, good renaming for the components.

CliveGlitch (approved):

FemboyDeveloper:
> Not so much to say, good renaming for the components.

yeah, felt like maybe I should have just committed but wanted to make sure you agreed.

PR #157, opened 2023-06-16
Fixing navbar bug

leekahung:
Very small thing, but would be important for UX. When selecting links from "GAME", it toggles burger menu to open. While we don't really see this if the user remains on desktop, it'll appear once we shrink the screen.

https://github.com/WarsWorld/WarsWorld/assets/14917816/24bdc554-ab1a-45de-b036-881b306e30f3

It should really only be opened when toggling the burger menu. So this PR should fix that problem. See below.

https://github.com/WarsWorld/WarsWorld/assets/14917816/e1939f6e-c4e8-4bec-9230-c53139c63df0


FemboyDeveloper (approved):
LGTM! Good job!

PR #159, opened 2023-06-16
Ltbackup/match card multi

LtBackup:
changes layout of match card to account for more than 2 players in a match. Also refines some enums and image names


FemboyDeveloper:
image
image
image
image

@LtBackup  I'm getting what seem some minor errors when I hard refresh the page. Not sure if this is just on me? I tried restarting my setup but the issue seems to continue.

Everything looks pretty good! Really like your work, only thing I'd add is randomizing names since I worry about long names like "Operation Mechstorm" but other than that, as always, love the clean designs!!!

FemboyDeveloper (approved):
LGTM! Keep up the great work!

PR #174, opened 2023-07-04
Nations Enum changes and Leaderboards Touch Ups

CliveGlitch:
Made some touch ups to the leaderboards page to be more responsive with larger monitors and used the enums from the enums file to clean up all the leaderboard files.

Reduced the Nations enums to just have one to simplify things, now you can get colors, text, icons or the nation itself with that enum. I've also changed all files that were affected by those changes.

image
image


JaviTrek (approved):
LGTM!

PR #175, opened 2023-07-04
Type safety fixes

CliveGlitch:
Quick PR for typescript type fixes, you were sending an array OR a single object from Unit Data, so that's why Typescript behaved weird, the getJSON type can be replaced by any or { [key: string]: any }, but it seems { [key: string]: string[] } is the best for type safety.

The map render still works as good.


JaviTrek (approved):
LGTM!

PR #180, opened 2023-07-07
Menu update

JaviTrek:
# The update

https://github.com/WarsWorld/WarsWorld/assets/96269542/d7ac5b60-1b13-4e1b-8d5e-001413d2623d

Basically menus now only leave when clicking outside a menu (instead of hovering out), menus are removed if another menu appears (no more double menus) and now menus cannot appear below the map, they will instead move upwards to fit tightly and smoothly (no more scrolling down to build units).

Also fixed some minor animations and made the menus smaller.

### DIFF
--- a/src/pages/match/[matchId].tsx
+++ b/src/pages/match/[matchId].tsx
@@ -15,6 +15,7 @@ import {
   SCALE_MODES,
   Sprite,
   Spritesheet,
+  Texture,
 } from "pixi.js";
 import { useEffect, useRef, useState } from "react";
 import { PlayerInMatch } from "shared/types/server-match-state";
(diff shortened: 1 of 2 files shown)


CliveGlitch:
Width and height will cut some maps depending on the screen size, everything else looks good.
image
image

src/pages/match/[matchId].tsx
```diff
@@ -60,23 +61,25 @@ const Match = ({ spriteData }) => {
...
-      width: 1400,
-      height: 2000,
+      width: window.outerWidth * 0.975,
+      height: window.outerHeight * 0.975,
       resizeTo: undefined,
     });
 
     //TODO: Button with + and - to change the scale of our stage, also needs
     // to have app.resize() working so we can resize the size of our app.
     app.stage.scale.set(2.6, 2.6);
```
Comment (CliveGlitch) [inline src/pages/match/[matchId].tsx:81]: Probably we're going to need a formula to set the size depending on the window screen and the map screen, maybe there is a better solution in pixijs documentation.

Comment (JaviTrek) [reply src/pages/match/[matchId].tsx:81]: Yeah, right now the map size is okay, just a minor issue that makes so if I do 100% screen width, it has an issue/I also want to find ways to resize it automatically

PR #181, opened 2023-07-09
Plus minus buttons, center game-map

JaviTrek:
## The update in a video.
https://github.com/WarsWorld/WarsWorld/assets/96269542/16db62bb-22ec-4125-9b75-98d453726139

### Important changes
Now pixi will only display the map/game, rather than be fullscreen. Before I was struggling to manage increasing the size of the game and not have any of it be off-screen and unreachable (map images increase, but the size of canvas (the html element) would not).

However this is fixed by "re-initializing" our PIXI.Application to increase its width and height instead. So now we can only use PIXI for game related things and keep every other component on HTML/CSS which will make it easier to work with (its no fun trying to style things in pixi...)


CliveGlitch (approved):
LGTM, just an observation, the scrollbar will not work in firefox and any other no chromium browser.

JaviTrek:
> LGTM, just an observation, the scrollbar will not work in firefox and any other no chromium browser.

thanks for the quick review @CliveGlitch ! Yeah the scroll bar was more of a test because I originally thought that it was causing an issue when I was using PIXI at full screen but since thats not the case anymore, But dully noted! Later I'll try looking for solutions for a better scroll bar across browsers

PR #190, opened 2023-10-11
Layout Overhaul

cliveglitch:
It started as an overhaul for the home page and now I got a lot of changes, mostly visuals to the whole layout.
- More responsive Footer.
image

- Navbar has been overhauled with a new way to open the hidden menu of games and a new look. I've found the unit images besides the navbar buttons to be out of place, but I can rollback if like it that way.
image
image

- Remade the Home Banner, and changed some buttons.
image
image

- New Title component that is being used on the News, How To Play,  Leaderboards and Your Matches pages.
image
image
image

-I've changed the buttons, selects and titles of the Your Matches page.
image

-Your matches page is the only page that I've also made some changes beside the visual stuff, I've divided the page into Create Match and Match Section components.
image
 
There are some more little changes I've made to these pages to be more responsive on other devices.


onetrickwolf (approved):
LGTM!

JaviTrek:
The changes look really nice, keep it up!

src/frontend/components/Navbar/NavItem.tsx
```diff
@@ -12,25 +13,22 @@ interface Props {
...
-        onClick={handleBurgerMenu}
-      >
-        {text}
-      </Link>
-      <img
+    <Link href={location} onClick={handleBurgerMenu}>
+      <div className="@flex @justify-center @items-center @gap-2 @h-full hover:@scale-[1.025]">
+        <NavButton>{text}</NavButton>
+
+        {/*<img
```
leekahung, line 26: Think we can remove the `img` if it's no longer in use.

leekahung (approved):
Other than the comment, think it looks good.

src/frontend/components/Navbar/NavItem.tsx
CliveGlitch: Agree, I think I've left it just in case.

PR #195, opened 2023-11-15
Basic Authentication

cliveglitch:
# Authentication Pull Request
I think I got a lot of progress with the authentication, and I wanted to receive some feedback about what I have.

## Features:
- Login/Signup modal.
- You can Login or Signup with credentials.
- You can Login or Signup with Discord, Google and GitHub providers if you fill the clientId and secret in the .env file for each provider.
- If you create an account by any means, like first login in with a provider or signing up, next auth will create a default player with the name of the user.
- The front end should tell you why you couldn't login or signup if you made a mistake, like wrong user or password, validations, if there is already a user with that email, etc.
- "your-games" will always default to the logged in user's first player.
- You can create games with any new user.
- Protected Page wrapper that will redirect you to the login page if you aren't logged in, example the your-games page.
- Passwords are saved as salted hashes with bcrypt.

## Does not have:
- Sessions.
- Forgot password functionality.
- Role based authorization.

image
image
image
image
image


next.config.js
```diff
@@ -4,3 +4,8 @@ module.exports = {
   eslint: { ignoreDuringBuilds: !!process.env.CI },
   reactStrictMode: false,
 };
+module.exports = {
+  env: {
+    NEXTAUTH_URL: process.env.NEXTAUTH_URL,
```
FunctionDJ, line 9: why only this one? shouldn't next auth read the environment variables regardless?

prisma/schema.prisma
```diff
@@ -39,12 +39,37 @@ enum Role {
...
+  email         String?   @unique
+  emailVerified DateTime?
+  image         String?
+  createdAt     DateTime  @default(now())
+  updatedAt     DateTime  @updatedAt
+  identities    Player[]
+  accounts      Account[]
+}
+
+model Account {
```
FunctionDJ, line 56: if a user uses both discord and email to login, they will have 2 accounts but 1 user on the DB, right?
how would you match them up?

prisma/schema.prisma
```diff
@@ -39,12 +39,37 @@ enum Role {
...
+  access_token      String? @db.Text
+  expires_at        Int?
+  token_type        String?
+  scope             String?
+  id_token          String? @db.Text
+  session_state     String?
+
+  user User @relation(fields: [userId], references: [id], onDelete: Cascade)
+
+  @@unique([provider, providerAccountId])
```
FunctionDJ, line 72: intuitively, i'd say the unique should be on `[provider, userId]`, no? and then do we even need the `id` anymore when these entrys can and need to be idenfitied by those two properties/columns?

prisma/schema.prisma
```diff
@@ -88,7 +113,7 @@ model Player {
   secret      Boolean       @default(false) // secret = true means that parent user is hidden on profile
   ChatMessage ChatMessage[]
 
-  User   User   @relation(fields: [userId], references: [id])
+  user   User   @relation(fields: [userId], references: [id], onDelete: Cascade)
```
FunctionDJ, line 116: please explain. to me this looks dangerous and would attempt to delete the user when the player gets deleted. or is it the other way around?

src/frontend/context/players.tsx
```diff
@@ -1,41 +1,67 @@
...
       {children}
     </playersContext.Provider>
   );
 };
 
 export const usePlayers = () => {
-  // TODO make sure development_player0 is set by default - currently not.
-  const ownedPlayers = useContext(playersContext);
+  const user = useContext(playersContext);
+  const ownedPlayers = user?.ownedPlayers;
```
FunctionDJ, line 49: have you implemented a "select default player" mechanic like the removed comment said on L25?

src/pages/api/WarsWorldAdapter.ts
```diff
@@ -0,0 +1,86 @@
+// Based on Prisma Adapter
+
+import type { PrismaClient } from "@prisma/client";
+import type { Adapter, AdapterAccount } from "@auth/core/adapters";
+import { AdapterUser } from "next-auth/adapters";
+import { Awaitable } from "next-auth";
+
+export default function WarsWorldAdapter(p: PrismaClient): Adapter {
```
FunctionDJ, line 8: can you please explain how this works and why it needs an adapter? (basically, pls explain next-auth lol sorry)

src/pages/api/authOptions.ts
```diff
@@ -0,0 +1,115 @@
...
+            email: dbUser.email,
+          };
+          return dbUserWithoutPassword as User;
+        }
+
+        return null;
+      },
+    }),
+    GithubProvider({
+      clientId: process.env.GITHUB_CLIENT_ID as string,
```
FunctionDJ, line 67: how does this handle when you don't specify the necessary environment variables? will there be errors? does the login option still show up?

src/pages/api/jwt.ts
```diff
@@ -0,0 +1,14 @@
+// This is an example of how to read a JSON Web Token from an API route
+import { getToken } from "next-auth/jwt";
+
+import type { NextApiRequest, NextApiResponse } from "next";
+
+export default async function handler(
```
FunctionDJ, line 6: i'd prefer not using `default` to incentivise using the actual name of this function, which probably should be something other than `handler`. or if it's truly just an example, the file name should be different.

src/pages/api/jwt.ts
```diff
@@ -0,0 +1,14 @@
...
+  req: NextApiRequest,
+  res: NextApiResponse
+) {
+  // If you don't have the NEXTAUTH_SECRET environment variable set,
```
FunctionDJ, line 10: when would this case come up?

src/server/hashPassword.ts
```diff
@@ -0,0 +1,10 @@
+import { genSalt, hash } from "bcrypt";
+
+const SALT_ROUNDS = 12;
+
+export default async function hashPassword(password: string) {
```
FunctionDJ, line 5: no `default` pls

src/server/routers/match/create.ts
```diff
@@ -6,18 +6,21 @@ import { prisma } from "server/prisma/prisma-client";
...
 import { PlayerInMatch } from "shared/types/server-match-state";
+import { withPlayerNameSchema } from "server/trpc/middleware/player";
 
 export const createMatchProcedure = playerBaseProcedure
   .input(
     withMapIdSchema.extend({
       selectedCO: coSchema,
     })
   )
+  .input(withPlayerNameSchema)
```
FunctionDJ, line 17: these schemas appear to be used only once, so it's not useful to extract them into reusable variables yet. also you're extending one schema and then using a second `.input` for another one, kind mix-and-matching. i think the mapId, selectedCO and playerName can just be a part of a single inline schema in this case.

src/server/schemas/auth.ts
```diff
@@ -0,0 +1,16 @@
+import { z } from "zod";
+
+// TODO: better password zod validation
+
+export const loginSchema = z.object({
+  username: z.string().min(1),
+  password: z.string().min(4).max(20),
```
FunctionDJ, line 7: we *can* add better password validation here, but i think it's simpler to just do that in the procedure layer. also i think the range should be something more like `8-40`.

src/server/schemas/auth.ts
```diff
@@ -0,0 +1,16 @@
...
+});
+
+export const signUpSchema = loginSchema.extend({
+  email: z.string().email(),
+  confirmPassword: z.string().min(4).max(20),
```
FunctionDJ, line 12: `confirm`? on the backend, this would just be `password` to us. the confirm part is only frontend.

src/server/schemas/auth.ts
```diff
@@ -0,0 +1,16 @@
...
+});
+
+export type ILogin = z.infer<typeof loginSchema>;
```
FunctionDJ, line 15: unless other types have been prefixed with `I`, it'd be consistent to remove them.

src/server/trpc/middleware/player.ts
```diff
@@ -68,10 +71,7 @@ export const playerMiddleware = t.middleware(async ({ ctx, next, input }) => {
 
 export const playerWithoutCurrentMiddleware = t.middleware(
   async ({ ctx, next }) => {
-    const ownedPlayers =
-      process.env.NODE_ENV === "development"
-        ? await getDevelopmentModeUserPlayers()
-        : await getProductionModeUserPlayers(ctx.session);
+    const ownedPlayers = await getProductionModeUserPlayers(ctx.session);
```
FunctionDJ, line 74: so there's no getDevModePlayers anymore? if the function handles this internally, the name should be updated. otherwise, dev mode players are missing now.

src/server/trpc/trpc-init.ts
```diff
@@ -1,10 +1,20 @@
...
-  errorFormatter({ shape }) {
-    return shape;
+  errorFormatter({ shape, error }) {
+    return {
+      ...shape,
+      data: {
+        ...shape.data,
+        zodError:
+          error.code === "BAD_REQUEST" && error.cause instanceof ZodError
+            ? error.cause.flatten()
```
FunctionDJ, line 15: i know this is from the trpc docs but some de-nesting could improve readability here.

FunctionDJ (changes requested):

src/pages/api/authOptions.ts
cliveglitch: You simply cannot log in with those providers. You don't need to specify them all, if you want to try to log in with GitHub, just fill GitHub's env variables, and if you want to log in with email and password you don't need to fill any.

prisma/schema.prisma
cliveglitch: Next Auth looks for the tables called User and Account to login and signup a user and that's why the naming, Accounts are all the providers the user have tied, like Discord, GitHub and Google. 

For now there is no functionality to bind two emails to the same user at the moment, that means if you try to sign up with email/password and then you try to login for the first time with discord it will throw an error. Also email signup is not made into an account, it just fills the password so the user can login with a password.

In short, you sign up as one user, that user can have cero to many accounts (like discord), and that user can have 1 to many players, players being the entities that can play the game.

prisma/schema.prisma
cliveglitch: I've took this schema from NextAuth's documentation: https://next-auth.js.org/v3/adapters/prisma
And I think it is what NextAuth expects for this to work.

prisma/schema.prisma
cliveglitch: Deletes the player when the user is deleted, I just put that there to make it more convenient to delete users and test the functionality, is it fine?

src/frontend/context/players.tsx
cliveglitch: Yes, even if you create a new User, when you enter the page the first player will be selected.

next.config.js
cliveglitch: I've continued Matthew's work and that is something he left there, it stills works fine without it but I cannot say why it's needed there.

src/pages/api/jwt.ts
cliveglitch: I understand, didn't had that on mind while trying to create my functions, I just added the default to the main function of the file, but I follow you.

src/server/trpc/middleware/player.ts
cliveglitch: Ok, change the name, I used only getProductionModeUserPlayers so I can see the players of any new user that doesn't have that prefix.

src/pages/api/jwt.ts
cliveglitch: Matthew's file, I think it'd come up with sessions, but for now I plan to leave with sessions in a future PR, so I'm inclined to delete this file.

src/server/routers/match/create.ts
cliveglitch: Doesn't seem like the name input is needed in the first place, I think I left it there when trying to show the player's name instead of their Id on those match cards but kinda forgot.

src/server/schemas/auth.ts
cliveglitch: hehe I was kinda playing with zod and what I can do, the latest commit should move the confirm check to the front.

src/server/trpc/trpc-init.ts
cliveglitch: mmm... how would you de-nest it?

src/server/hashPassword.ts
cliveglitch: Done

src/server/schemas/auth.ts
cliveglitch: Done

next.config.js
FunctionDJ: then it should be removed.

prisma/schema.prisma
FunctionDJ: it's probably fine, but if a user gets deleted, we need to do cleanups regardless.
players can have clan relationships, which need to be resolved. the "ownedClans" one can't be cascade-deleted i think.
i'm not sure if we should use cascade deletion or just do everything manually because we can't fully "automate" it anyway.

src/frontend/components/auth/SignupForm.tsx
```diff
@@ -22,6 +22,8 @@ export default function SignupForm({
     confirmPassword: "",
   });
 
+  const [doPasswordsMatch, setDoPasswordsMatch] = useState(true);
```
FunctionDJ, line 25: i don't think state is needed. just do `passwordsMatch = signupData.password === signupData.confirmPassword`

src/pages/api/authOptions.ts
FunctionDJ: i've tested it and the buttons show up and i think clicking them will start the oauth flow which will immediately just redirect to the application (like a refresh) when these environment variables are are missing/invalid.

i think it would be a good idea if we made this more explicit in the GUI (e.g. for new developers) where the buttons are greyed out or just removed if the credentials are not set.

the latter might be easy to achieve by padding the providers conditionally like we've done in the past at points.
what do you think?

src/server/trpc/trpc-init.ts
FunctionDJ: i think pulling the value of `zodError` into a variable makes sense.

src/pages/api/WarsWorldAdapter.ts
FunctionDJ: so when someone uses email/password to sign up, they don't get a default player, right?
can we do something about that?

cliveglitch: They do, and I already have that done, you just need to add that logic separately.
NextAuth is not so keen with credentials login with password, so they design it to take more effort to set it up purposefully.

src/pages/api/authOptions.ts
cliveglitch: sounds like a good idea, I'm gonna work on it, later I'll tell you when it's done.

prisma/schema.prisma
cliveglitch: would need some time to investigate this, for now I got this for testing purposes so we could change if it has no real use.

cliveglitch: I opted to remove the the onCascade deletion for now, if I can find a way to automatically delete a user completely then we can discuss that later.

src/pages/api/authOptions.ts
cliveglitch: Done, it now greys out the buttons when you don't have the necessary env variables for that provider. 
In this example I don't have the env for Google's provider.
image

src/server/trpc/trpc-init.ts
cliveglitch: Done

src/frontend/components/auth/SignupForm.tsx
cliveglitch: Done

src/frontend/components/modals/LoginSignupModal.tsx
```diff
@@ -17,13 +23,29 @@ interface Props {
 }
 
 export default function LoginSignupModal({ isOpen, setIsOpen, width }: Props) {
+  const posibleProviders = ["github", "discord", "google"];
+  const [currentProviders, setCurrentProviders] = useState<Record<
+    LiteralUnion<BuiltInProviderType, string>,
```
FunctionDJ, line 28: maybe this type expression can be simplified with something like
```ts
Awaited<ReturnType<typeof getProviders>>
```
but i'll leave this up to you if it's better or not (if it works)

src/frontend/components/modals/LoginSignupModal.tsx
```diff
@@ -47,6 +69,21 @@ export default function LoginSignupModal({ isOpen, setIsOpen, width }: Props) {
 
   const onClose = async () => await setIsOpen(false);
 
+  const renderProviders = () =>
```
FunctionDJ, line 72: inline components should be avoided because of re-renders (and it would also break hooks/state). just make `currentProviders` a prop. you don't need to turn `possibleProviders` into a prop since it can live outside of components anyway.

src/pages/api/authOptions.ts
```diff
@@ -13,60 +13,101 @@ import { Adapter } from "next-auth/adapters";
...
-          };
-          return dbUserWithoutPassword as User;
-        }
+    },
+    async authorize(credentials) {
+      if (!credentials || !credentials.name || !credentials.password) {
+        return null;
+      }
+
+      const creds = await loginSchema.parseAsync({
```
FunctionDJ, line 62: this code could be simplified by making the submitted data fit to the loginSchema or vice versa.
then you could do `const { username, password } = loginSchema.safeParse(credentials)`
([don't use `async` unless the schema uses async refiners etc.](https://github.com/colinhacks/zod#parseasync))
(also i think [.strict()](https://github.com/colinhacks/zod#strict) would be a good idea here to ban needless properties which could indicate issues, up to you whether to apply it here when used or to put it on `loginSchema`'s definition directly)

src/pages/api/authOptions.ts
```diff
@@ -13,60 +13,101 @@ import { Adapter } from "next-auth/adapters";
...
-      },
-    }),
+      }
+
+      const doPasswordsMatch = await compare(
+        credentials.password,
+        dbUser.password
+      );
+
+      if (doPasswordsMatch) {
```
FunctionDJ, line 80: i think code makes more sense with early-returns in "failure cases", related to the "fail fast" principle.
in practise, that'd mean `if (!doPasswordsMatch) { return null; } ...`

src/pages/api/authOptions.ts
```diff
@@ -84,12 +125,14 @@ export const authOptions: NextAuthOptions = {
...
-      clientSecret: process.env.GOOGLE_CLIENT_SECRET as string,
-    }),
-  ],
+    })
+  );
+
+export const authOptions: NextAuthOptions = {
+  secret: process.env.NEXTAUTH_SECRET,
+  adapter: adapter,
+  debug: true,
```
FunctionDJ, line 134: maybe something like `debug: process.env.NODE_ENV === "dev"` ?
idk which env vars are currently available to check for development.
maybe just `console.log(process.env)` and see if something's in there.

FunctionDJ (changes requested):

src/frontend/components/modals/LoginSignupModal.tsx
```diff
@@ -17,13 +23,29 @@ interface Props {
...
   const searchParams = useSearchParams();
   const router = useRouter();
 
   const [didSignUp, setDidSignUp] = useState(false);
+
   const isSignupForm = searchParams.has("SignUpForm");
   const callbackUrl = searchParams.get("callbackUrl");
 
+  useEffect(() => {
+    const setProviders = async () => {
```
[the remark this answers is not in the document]
cliveglitch: Much simpler.

src/frontend/components/modals/LoginSignupModal.tsx
cliveglitch: It works, so it's worth it to change it.

src/pages/api/authOptions.ts
cliveglitch: Done.

src/pages/api/authOptions.ts
cliveglitch: Decided to leave out the .strict(), but everything else is done.

src/frontend/components/modals/LoginSignupModal.tsx
cliveglitch: I've just put back the map into the jsx

src/frontend/components/modals/LoginSignupModal.tsx
```diff
@@ -17,13 +23,29 @@ interface Props {
 }
 
 export default function LoginSignupModal({ isOpen, setIsOpen, width }: Props) {
+  const posibleProviders = ["github", "discord", "google"];
```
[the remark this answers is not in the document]
cliveglitch: Done.

src/frontend/components/auth/SignupForm.tsx
```diff
@@ -0,0 +1,140 @@
...
+    user: "",
+    email: "",
+    password: "",
+    confirmPassword: "",
+  });
+
+  const doPasswordsMatch =
+    signupData.password === signupData.confirmPassword ||
+    signupData.password === "" ||
+    signupData.confirmPassword === "";
```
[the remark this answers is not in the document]
cliveglitch: I think it's a misunderstanding, I've changed the code a bit to make it more obvious, but the purpose of that conditional is when to `show` an error, the logic is, when the passwords do not match and there is at least a character on both the inputs then it will show the error.
It's to not bother the user with a "Passwords do not match message when they are typing the first password".

src/frontend/components/auth/SignupForm.tsx
```diff
@@ -38,12 +43,15 @@ export default function SignupForm({
   const onSubmitSignupForm = async (event: FormEvent) => {
     event.preventDefault();
 
+    if (signupData.confirmPassword !== signupData.password) {
```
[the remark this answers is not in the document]
cliveglitch: You can submit when passwords do not match, there is no code that stops the user not to submit until that point.
The code on the review above was to show the error.

src/frontend/components/auth/SignupForm.tsx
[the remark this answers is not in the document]
FunctionDJ: Ah, i get it now. Please change the variable to something like
```ts
const showPasswordsDoNotMatch = signupData.password !== "" && signupData.confirmPassword !== "" && signupData.password !== signupData.confirmPassword
```
This makes the logic and the intent behind this code much easier to understand, because the variable isn't exclusively about whether the passwords match, but also if it makes sense to show a "passwords don't match" hint.

src/frontend/components/auth/SignupForm.tsx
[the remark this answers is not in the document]
FunctionDJ: Why not disable the submit button when passwords don't match instead?

src/frontend/components/modals/LoginSignupModal.tsx
FunctionDJ: yes, that's fine.

src/pages/api/authOptions.ts
```diff
@@ -55,37 +55,37 @@ const providers: Provider[] = [
...
-        return dbUserWithoutPassword as User;
+      if (!doPasswordsMatch) {
+        return null;
       }
 
-      return null;
+      return {
+        name: dbUser.name,
+        email: dbUser.email,
+      } as User;
```
FunctionDJ, line 88: What's up with this type override?

src/frontend/components/auth/SignupForm.tsx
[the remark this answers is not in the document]
cliveglitch: Kinda does the same I suppose, you just get rejected after you click it. I've change it no problem, and also added so you cannot press the button if you have empty fields because it doesn't make sense to disable it only with the passwords.

src/frontend/components/auth/SignupForm.tsx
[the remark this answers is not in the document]
cliveglitch: Done.

src/pages/api/authOptions.ts
cliveglitch: You get this error otherwise:
image

You can add the id to the object (I did that on the last commit), and it works the same.

FunctionDJ (approved):

PR #203, opened 2023-12-23
Fix errors with frontend

JaviTrek:
There were some small errors (see image below) involving if there wasnt a teamMapping in a match and if there were no units in a match (or well, at least if units property wasn't defined in the match).

image

These small fixes allow the frontend to get the current player via usePlayers() useContext implementation. Which therefore, allows devs to work on match creation.

# Important Note
If these fixes are already addressed in some work in progress you have function/pau, I appreciate if we could merge this anyways because its infinitely small but it allows us to effectively get the currentPlayer and other players in the frontend effectively, thanks!


JaviTrek:
Did a smaller fix and pushed to main.

PR #219, opened 2024-01-15
Added a context in your-matches page

### LINKED ISSUE #214: Eliminate location.reload() on MatchCardSetup (use state or refetching matches instead)
### Where is the issue
In `src/frontend/components/match/matchCardSetup.tsx`, you'll see a couple location.reload() sprinkled around here and there

### What is the issue
So `location.reload() `triggers a refresh of the current window, this creates a slower and worse user experience besides being a bad practice. These location.reloads are just there because we need to fetch new data again so what we really need to do is simply callback the matches router or possibly just update the frontend state to match the changes.

They are there because its easier to join a match and just refresh the page so we re-query matches rather than actually using state smartly. I implemented them because I wanted to have stuff "just working" rather than being optimal.

rjam13:
Closes #214 

This is done so that refetchYourMatches and refetchAllMatches can be called way down the component tree.

The result of this allows us to join, create, and leave games without having to reload the entire page.

### DIFF
--- a/src/frontend/components/match/CreateMatch.tsx
+++ b/src/frontend/components/match/CreateMatch.tsx
@@ -4,22 +4,20 @@ import Select from "frontend/components/layout/Select";
 import SquareButton from "frontend/components/layout/SquareButton";
 import { usePlayers } from "frontend/context/players";
 import { trpc } from "frontend/utils/trpc-client";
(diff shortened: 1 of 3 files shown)


JaviTrek:
@rjam13 great job! Keep it up <3

Issue #213, opened 2024-01-07
Add Comments to Articles


rjam13:
This issue is blocked by #212 . 

Will be triaged soon

evayde:
Hi, I'd implement that. 

Should it go directly into the article or should we create a new area below the article that holds the comments?

rjam13:
@evayde 
If we are talking about visually how it'd look, I'd prefer a new area below. 
Design isn't the biggest concern at the moment so feel free to experiment and ask the discord for feedback.

Be sure to link a pull request to this issue.

JaviTrek:
@evayde pls join our Discord to keep up with updates and stuff, dont want your work to conflict w someone elses https://discord.gg/9cgTs5ZGT2

evayde:
I added a simple mechanic to display that the user should be logged in to comment. This could, of course, be a lot prettier. 
We should add follow up tickets for the following tasks:

- Comment thumbs up/down
- Replies to comments
- Rich Editor for comments (with Emojis and so on)
- Comment design

PR #222, opened 2024-01-18
Feature/article comments

evayde:
https://github.com/WarsWorld/WarsWorld/issues/213


rjam13:
LGTM! @cliveglitch what do you think?

